PR #5, opened 2025-01-10
VIDCS-3227: Room name with + did not show participants

behei-vonage:
#### What is this PR doing?

This PR fixes an issue where room name with + did not show participants. The issue was caused because in some cases we used our own regex to determine room validity, and in some cases we used a built-in `encodeURIComponent` which was more strict in some cases. 

This PR moves out room validation to its own helper and uses that instead of `encodeURIComponent.` It also adds on a few test cases for room names. 

To reproduce the issue, checkout the `main` branch.
Try joining the room `localhost:5173/waiting-room/me+me`, proceed to enter the room and notice that the URL in the address bar changes `+` to `%20`.
Try adding another user go through the same path, notice that even though URL looks the same, you end up in split rooms.

To reproduce the fix, checkout this branch.
Do the steps mentioned above and notice that you no longer see the issue.

Resolves [VIDCS-3227](https://jira.vonage.com/browse/VIDCS-3227)

[👎 ] Resolves a `Known Issue`.
[👎 ] Resolves an item reported in `Issues`.
[ ] If yes, did you close the item in `Issues`?

### DIFF
--- a/frontend/src/components/RoomNameInput/RoomNameInput.spec.tsx
+++ b/frontend/src/components/RoomNameInput/RoomNameInput.spec.tsx
@@ -12,22 +12,78 @@ describe('RoomNameInputComponent', () => {
     setHasError = vi.fn();
   });
 
-  it('should set the room name if the input is valid', () => {
-    render(
-      <RoomNameInput
-        setRoomName={setRoomName}
-        setHasError={setHasError}
-        roomName="test-room"
-        hasError={false}
-      />
-    );
+  describe('should set the room name', () => {
+    const testCases = [
+      { input: 'test_room' },
+      { input: 'test+room' },
+      { input: 'another-test_room' },
+      { input: '123testroom' },
+    ];
 
-    const input = screen.getByRole('textbox');
-    fireEvent.change(input, { target: { value: 'new-test-room' } });
+    testCases.forEach(({ input }) => {
+      it('if the input is valid', () => {
+        render(
+          <RoomNameInput
+            setRoomName={setRoomName}
+            setHasError={setHasError}
+            roomName="test-room"
+            hasError={false}
+          />
+        );
 
-    expect(setHasError).toHaveBeenCalledWith(false);
-    expect(setRoomName).toHaveBeenCalledWith('new-test-room');
+        const inputBox = screen.getByRole('textbox');
+        fireEvent.change(inputBox, { target: { value: input } });
+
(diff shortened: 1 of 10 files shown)


frontend/src/components/RoomNameInput/RoomNameInput.spec.tsx
```diff
@@ -12,23 +12,77 @@ describe('RoomNameInputComponent', () => {
...
+
+        const inputBox = screen.getByRole('textbox');
+        fireEvent.change(inputBox, { target: { value: input } });
+
+        expect(setHasError).toHaveBeenCalledWith(true);
+        expect(setRoomName).not.toHaveBeenCalledWith(input);
+      });
+    });
+  });
+  //   render(
```
v-kpheng, line 70: Looks like some stray code here :-)

frontend/src/utils/validateRoomName/validateRoomName.ts
```diff
@@ -0,0 +1,7 @@
+const validateRoomName = (name: string) => {
```
v-kpheng, line 1: This is predicate so maybe rename to `isFoo` instead?

frontend/src/components/WaitingRoom/UserNameInput/UserNameInput.tsx
```diff
@@ -67,7 +68,10 @@ const UsernameInput = ({ username, setUsername }: UserNameInputProps): ReactElem
   const handleJoinClick = (event: MouseEvent) => {
     event.preventDefault();
     if (validateForm() && roomName) {
-      const sanitizedRoomName = encodeURIComponent(roomName);
+      const validatedRoomName = validateRoomName(roomName);
+      if (!validatedRoomName) {
+        return;
+      }
```
v-kpheng, line 74: Nit: probably more readable with some light refactoring.

```
if (!isValidRoomName(roomName)) {
  return;
}
```

v-kpheng: Does `encodeURIComponent` have any value?  I think it might be used elsewhere, yes?

v-kpheng (commented):
Looks good! 💪 

That said, left some questions/comments.  Thanks! 🙏

maikthomas (approved):
LGTM!

DeliaTok:
looks good!

PR #18, opened 2025-01-20
VIDCS-3202: Fix room creation race condition - split room / session

maikthomas:
#### What is this PR doing?
There is a race condition in the backend where two simultaneous requests can overlap due to the async nature of the code. In this case the 2nd request can end up creating a 2nd room and overwriting the 1st room in storage. This leaves the first person in a room that is now inaccesible and the 2nd and future participants join the 2nd session.

- Running two curl (or more) commands joined with an `&` operator to run in parallel
- Try reproing the issue first on `main` then checkout this branch to test the fix
- curl:
```shell
curl 'http://localhost:3345/session/some-new-session'  & curl 'http://localhost:3345/session/some-new-session'
```
Remember that one you've used a room name it will be stored in memory, this will only reproduce the first time.

Resolves [VIDCS-3202](https://jira.vonage.com/browse/VIDCS-3202)

[ ] Resolves a `Known Issue`.
[ ] If yes, did you close the item in `Issues`?


backend/helpers/__mocks__/config.ts
```diff
@@ -0,0 +1,16 @@
...
+      return {
+        apiKey: 'test-api-key',
+        apiSecret: 'test-api-secret',
+        applicationId: 'test-application-id',
+        privateKey: 'test-private-key',
+        provider: 'opentok',
+      };
+    }),
+  };
+};
```
maikthomas, line 15: Moved this here since I was using it twice

backend/helpers/blockCallsForArgs.ts
```diff
@@ -0,0 +1,26 @@
...
+      const promiseWithResolvers = Promise.withResolvers<null>();
+      callsInProgress[key] = promiseWithResolvers;
+    }
+    const res = await fn(key, ...args);
+    callsInProgress[key]?.resolve(null);
+    delete callsInProgress[key];
+    return res;
+  };
+};
+
```
maikthomas, line 25: This is the util I came up with the help us deal with async session creation / pulling from memory for requests that come in at the same time

backend/routes/session.ts
```diff
@@ -1,38 +1,46 @@
...
-        apiKey: data.apiKey,
-      });
-    }
+    const sessionId = await getOrCreateSession(roomName);
+    const data = videoService.generateToken(sessionId);
+    res.json({
+      sessionId,
+      token: data.token,
+      apiKey: data.apiKey,
+    });
```
maikthomas, line 36: I refactored this to:
- Split phase - separate session creation from token creation
- wrap getOrCreate session in the async blocker wrapper

backend/routes/session.ts
```diff
@@ -1,38 +1,46 @@
...
   } catch (error: unknown) {
     const message = error instanceof Error ? error.message : error;
     res.status(500).send({ message });
   }
 });
 
-sessionRouter.post('/:room/startArchive', async (req: Request, res: Response) => {
+sessionRouter.post('/:room/startArchive', async (req: Request<{ room: string }>, res: Response) => {
```
maikthomas, line 43: I edited the TSConfig to disable `noUncheckedIndexedAccess` because it was letting us access and use anything from an object without checking. After disabling this rule I had to add the types here and in the following route declarations

backend/routes/session.ts
```diff
@@ -52,22 +60,25 @@ sessionRouter.post('/:room/startArchive', async (req: Request, res: Response) =>
...
-    const { archiveId } = req.params;
-    if (archiveId) {
-      const responseArchiveId = await videoService.stopArchive(archiveId);
-      res.json({
-        archiveId: responseArchiveId,
-        status: 200,
-      });
+sessionRouter.post(
+  '/:room/:archiveId/stopArchive',
+  async (req: Request<{ room: string; archiveId: string }>, res: Response) => {
```
maikthomas, line 65: only change here is the addition of this type, the rest is just formatting

backend/tests/getOrCreateSession.test.ts
```diff
@@ -0,0 +1,134 @@
...
+          .fn<() => Promise<Archive[]>>()
+          .mockResolvedValue([{ id: 'archive1' }, { id: 'archive2' }] as unknown as Archive[]),
+      };
+    }),
+  };
+});
+
+const { getOrCreateSession } = await import('../routes/session');
+
+describe('getOrCreateSession', () => {
```
maikthomas, line 48: The reason this is not in session.test.ts is that those tests use supertest to test the express endpoints, but those requests do not run in parallel like in real life, even if you don't await them, they run one by one.

backend/tests/session.test.ts
```diff
@@ -3,20 +3,9 @@ import request from 'supertest';
...
-        applicationId: 'test-application-id',
-        privateKey: 'test-private-key',
-        provider: 'opentok',
-      };
-    }),
-  };
-});
+await jest.unstable_mockModule('../helpers/config', mockOpentokConfig);
 
 const mockVcrSessionStorage = {
```
maikthomas, line 10: I actually believe a lot of this test suite to not be doing what it looks like its doing, but I don't want to do everything in this PR

backend/tsconfig.json
```diff
@@ -96,7 +96,7 @@
     // "exactOptionalPropertyTypes": true,               /* Interpret optional property types as written, rather than adding 'undefined'. */
     // "noImplicitReturns": true,                        /* Enable error reporting for codepaths that do not explicitly return in a function. */
     // "noFallthroughCasesInSwitch": true,               /* Enable error reporting for fallthrough cases in switch statements. */
-    // "noUncheckedIndexedAccess": true,                 /* Add 'undefined' to a type when accessed using an index. */
+    "noUncheckedIndexedAccess": false,                 /* Add 'undefined' to a type when accessed using an index. */
```
maikthomas, line 99: This was typing everything accessed via an index as defined, disabling for safer checking.

maikthomas (commented):
👨‍🍳

behei-vonage (approved):
LGTM! this is impressive! 🚀

backend/routes/session.ts
```diff
@@ -1,38 +1,46 @@
...
 
-sessionRouter.get('/:room', async (req: Request, res: Response) => {
+/**
+ * Gets or creates a session in a thread safe (async operation safe) way.
+ * The function is wrapped in blockCallsForArgs to ensure that simultaneous calls for a room name return the same session.
+ * Not blocking these calls would result in separate sessions being created for the same room name.
+ * @param {string} roomName - name of the meeting room
+ * @returns {Promise<string>} the sessionId
+ */
+export const getOrCreateSession = blockCallsForArgs(async (roomName: string) => {
```
cpettet, line 17: Is it worth moving this to its own helper file as a higher-order function/factory to tidy up the `session.ts` file?

cpettet (approved):
LGTM Great job! Left some nits and a question, but no blockers. :shipit:

backend/routes/session.ts
maikthomas: I could do, that would probably simplify the test file too with some injection of the services

backend/videoService/opentokVideoService.ts
```diff
@@ -17,17 +17,14 @@ class OpenTokVideoService implements VideoService {
...
       this.opentok.createSession({ mediaMode: 'routed' }, (error, session) => {
         if (error || !session) {
           reject(error ?? new Error('Unknown error occurred, no session created.'));
         } else {
           const { sessionId } = session;
-          const token = this.opentok.generateToken(sessionId, {
-            role: OpenTokVideoService.getTokenRole(),
-          });
-          resolve({ sessionId, token });
+          resolve({ sessionId });
```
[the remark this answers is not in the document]
maikthomas: Yea I actually meant to change this, thanks for spotting

behei-vonage (approved):
LGTM! 🚀

cpettet (approved):
LGTM Great job! :shipit:

dwivedisachin:
Tested LGTM!!

PR #73, opened 2025-02-14
VIDCS-3398: Setup local VCR deployment workflow

maikthomas:
#### What is this PR doing?
- Setup the repo for easy deployment to VCR from local env for developers

Please follow the instructions added in the readme and validate that everything works as expected.

Resolves [VIDCS-3398](https://jira.vonage.com/browse/VIDCS-3398)

#### Checklist
[👍] Branch is based on `develop` (not `main`).
[👎] Resolves a `Known Issue`.
[👎] If yes, did you remove the item from the `docs/KNOWN_ISSUES.md`? 
[👎] Resolves an item reported in `Issues`.


.vcrignore
maikthomas, line 1: This ignore avoid sending files we don't want to VCR, which increases the deploy time and is also pointless - also node_modules will not work there if built on mac.

backend/package.json
```diff
@@ -24,7 +24,8 @@
     "dotenv": "^16.0.3",
     "express": "^4.21.2",
     "form-data": "^4.0.1",
-    "opentok": "^2.16.0"
```
maikthomas, line 27: Just moving things around to make sure all dependencies required to build and run the project are in dependencies not devDependencies. By default VCR will run yarn in production mode, only installing dependencies.

package.json
```diff
@@ -15,6 +15,8 @@
   "scripts": {
     "build": "yarn workspace frontend build",
     "deploy-vcr": "yarn build && vcr deploy",
+    "deploy-vcr-local": "dotenv -- yarn deploy-vcr-local-with-env",
+    "deploy-vcr-local-with-env": "vcr deploy -f ./vcr_local.yml --app-id $VCR_APP_ID",
```
maikthomas, line 19: to use the dotenv values you need to use a separate command. If I put all this in one line $VCR_APP_ID would not be defined.

.vcrignore
maikthomas, line 1: This ignore avoid sending files we don't want to VCR, which increases the deploy time and is also pointless - also node_modules will not work there if built on mac.

.env.example
```diff
@@ -0,0 +1,2 @@
+# VCR Application ID for deploying to VCR from local machine
+VCR_APP_ID=
```
javiermolsanz, line 2: Question : Why are we calling this VCR_APP_ID rather than VONAGE_APP_ID?

package.json
javiermolsanz: So this means that if you want to deploy the app with the variables in your env file you need to run 
`yarn deploy-vcr-local`
correct?

maikthomas: Yes, the other way is to put this in a string in a bash command. 
Maybe we could rename this to something less user friendly

.env.example
maikthomas: VONAGE_APP_ID is already in the backend .env for using vonage unified video environment. I think it would be confusing, this specifies that this is for use by the VCR script.

vcrBuildLocal.sh
```diff
@@ -0,0 +1,5 @@
+ #!/bin/bash
```
javiermolsanz, line 1: A comment here would be nice. Indicating that this is to build dependencies and build the app in the machine used by VCR rather than uploading the build built locally https://developer.vonage.com/en/vonage-cloud-runtime/guides/manifest#build-script

javiermolsanz (approved):
I'm going to approve for now and once there's another ✅ I'll test it. Also we need to update the readme with the new deployment instructions

dwivedisachin:
I just tested it and I was able to deploy to my instance. 🚀
Do you think we should add the steps to create app id too 🤔

README.md
```diff
@@ -157,6 +158,22 @@ yarn start
...
+## Deployment to Vonage Cloud Runtime
+
+You can easily deploy your local branch to Vonage Cloud Runtime (VCR) using the tools in this repository.
+
+First, set up your VCR application at https://developer.vonage.com/en/vonage-cloud-runtime/overview. One you have your application, get your application ID  and set it in your config  `VCR_APP_ID` in the top level `./env` file.
+You should use a separate application to your video application to avoid issues with your private key.
+
+Then create a workspace named `dev` (Or change the instance name in `./vcr_local.yml#L4`). See https://developer.vonage.com/en/vonage-cloud-runtime/getting-started/managing-projects.
+
+Install the `vcr` cli globally with `npm i -g vcr`.
```
abdulajet, line 170: CLI is not installed via NPM
https://developer.vonage.com/en/vonage-cloud-runtime/getting-started/working-locally#cli-installation

README.md
```diff
@@ -157,6 +158,22 @@ yarn start
 
 The app and API are both served on  [http://localhost:3345/](http://localhost:3345/)
 
+## Deployment to Vonage Cloud Runtime
+
+You can easily deploy your local branch to Vonage Cloud Runtime (VCR) using the tools in this repository.
+
+First, set up your VCR application at https://developer.vonage.com/en/vonage-cloud-runtime/overview. One you have your application, get your application ID  and set it in your config  `VCR_APP_ID` in the top level `./env` file.
```
abdulajet, line 165: i think this is just a normal vonage application? if so it can be done via the CLI or the API dashboard. I recommend moving the CLI install first then just having a copy and paste for creating a new application

maikthomas: Understood, I have made this more generic and changed the link to point to the dashboard 🙏

README.md
```diff
@@ -157,6 +158,22 @@ yarn start
 
 The app and API are both served on  [http://localhost:3345/](http://localhost:3345/)
 
+## Deployment to Vonage Cloud Runtime
+
+You can easily deploy your local branch to Vonage Cloud Runtime (VCR) using the tools in this repository. See https://developer.vonage.com/en/vonage-cloud-runtime/overview for an overview of Vonage Cloud Runtime.
+
+Firstly, install the VCR cli: https://developer.vonage.com/en/vonage-cloud-runtime/getting-started/working-locally#cli-installation and run `vcr configure`.
+
```
javiermolsanz, line 166: I think we're missing one step which is running vcr init to create an instance and an application for VCR to use for deployment. Let's huddle later

javiermolsanz (changes requested):
Left one comment regarding vcr init https://developer.vonage.com/en/vonage-cloud-runtime/getting-started/working-locally#using-the-cli let's catch up later

dwivedisachin (dismissed):
LGTM!!

javiermolsanz (approved):

dwivedisachin (approved):
Also, tested LGTM!! 🚀

PR #86, opened 2025-02-20
VIDCS-3438: Update main branch for VERA 1.0.2

behei-vonage:
#### What is this PR doing?
Updates `main` with changes from `develop` aka `rc-1.0.2`

#### How should this be manually tested?
- CI passes

#### What are the relevant tickets?

Resolves [VIDCS-3438](https://jira.vonage.com/browse/VIDCS-3438)

[ ] Resolves a `Known Issue`.
[ ] If yes, did you remove the item from the `docs/KNOWN_ISSUES.md`? 
[ ] Resolves an item reported in `Issues`.
If yes, which issue? [Issue Number](https://github.com/Vonage/vonage-video-react-app/issues/)?
[ ] If yes, did you close the item in `Issues`?

PR #87, opened 2025-02-20
VIDCS-3439: Skip Auto-set Base Branch to Develop / set-base-branch (pull_request) Job in VERA during the release

behei-vonage:
#### What is this PR doing?

This PR adds some additional logic to enable skipping auto setting job so that when we have a release candidate branch merging into `main` during the release process, it doesn't fail the job like it did [here](https://github.com/Vonage/vonage-video-react-app/pull/86)

Example this actually working can be found here: https://github.com/Vonage/vonage-video-react-app/pull/88

Create a new branch with something that has `rc` in it, such as `rc-testing-vidcs-3439` from this pull request, make a small minor change somewhere in, for example, README.md.
Push up your branch to GitHub and create a new Pull Request.
Make sure that you are merging in your PR into the `main` branch.
Click "Create Pull Request"
Notice that the set auto base job did not fail. 
Don't forget to Close your test PR.

Example of my own testing can be found here: https://github.com/Vonage/vonage-video-react-app/pull/88

Resolves [VIDCS-3439](https://jira.vonage.com/browse/VIDCS-3439)

#### Checklist
[ ] Resolves a `Known Issue`.


.github/workflows/auto-set-base.yml
```diff
@@ -16,7 +16,18 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v3
 
+      - name: Check for "release-candidate" label
+        id: check-release-candidate
+        uses: actions/github-script@v6
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const pr = context.payload.pull_request;
```
v-kpheng, line 25: Does GH run javascript code?  If so, TIL :-)

behei-vonage: yup, we do this in JS SDK for some labels, too 🙏

maikthomas: Are we going to call these branches `rc-*` if so we could probably skip the label and this script and just edit the bash conditional to include that

behei-vonage: sure thing, I can do it that way, too. I updated it

.github/workflows/auto-set-base.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
         run: |
-          if [ "$TARGET_BRANCH" = "main" ] && [ "$BASE_BRANCH" != "develop" ]; then
+          if [[ "$TARGET_BRANCH" == "main" && "$BASE_BRANCH" != "develop" && ! "$BASE_BRANCH" == rc-* ]]; then
```
maikthomas, line 26: Is this `!` before the expression common for bash? rather than `!=` which we're using in the develop conditional?
If we can use `!=` instead I think we should for consistency.

maikthomas (commented):
One small question 🙏

.github/workflows/auto-set-base.yml
behei-vonage: good question! we have to use `!` before `"$BASE_BRANCH" == rc-*` since that's what used for pattern matching. In other conditionals we have here, like checking `main` and `develop` branch we can use the `==` and `!=` directly

maikthomas: I ran some tests locally and it seems `!=` will work fine for pattern matching too

maikthomas: https://www.namehero.com/blog/bash-string-comparison-the-comprehensive-guide/#3-3-partial-string-matching

behei-vonage: I think I was missing the `""` around `rc-*` when I was trying out the `!=.` it seems to work now with the quotes. thanks Mike!

.github/workflows/auto-set-base.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
         run: |
-          if [ "$TARGET_BRANCH" = "main" ] && [ "$BASE_BRANCH" != "develop" ]; then
+          if [[ "$TARGET_BRANCH" == "main" && "$BASE_BRANCH" != "develop" && "$BASE_BRANCH" != "rc-*" ]]; then
```
v-kpheng, line 26: Don't we want `"$BASE_BRANCH" != rc-*` instead?  Reason being, the quotes means it'll be a literal string match (i.e., "rc-*", instead of `rc-<SOME_VERSION>`).

I wrote a test script too:
```
#! /bin/bash

BASE_BRANCH="rc-2.19.2"
if [[ "$BASE_BRANCH" == rc-* ]]; then
    echo "matches"
else
    echo "no matches"
fi
```

v-kpheng (commented):
Looks good! 💪 

That said, left a question.  Thanks! 🙏

.github/workflows/auto-set-base.yml
behei-vonage: doing both approaches returns `matches.`

For example, the following:
```
BASE_BRANCH="rc-2.19.2"
if [[ "$BASE_BRANCH" != "rc-*" ]]; then
    echo "matches"
else
    echo "no matches"
fi
```
also returned `matches.` It's probably just a matter of preference at this point 😉

v-kpheng: I updated the script to the following
```
BASE_BRANCH="rc-2.19.2"
if [[ "$BASE_BRANCH" == "rc-*" ]]; then
    echo "matches"
else
    echo "no matches"
fi
```

It gives me the following result:
```
bash-3.2$ cat test.sh && ./test.sh
BASE_BRANCH="rc-2.19.2"
if [[ "$BASE_BRANCH" == "rc-*" ]]; then
    echo "matches"
else
    echo "no matches"
fi
no matches
bash-3.2$ 
```

If I change the expression (notice the lack of quotes on the right-hand side of the equality), it gives me the right result though:
```
bash-3.2$ cat test.sh && ./test.sh
BASE_BRANCH="rc-2.19.2"
if [[ "$BASE_BRANCH" == rc-* ]]; then
    echo "matches"
else
    echo "no matches"
fi
matches
bash-3.2$ 
```

behei-vonage: right but notice that the expression I use in the PR is the following:
```
"$BASE_BRANCH" != "rc-*" 
```
it uses `!=` instead of `==` 🙏

v-kpheng: Let's meet up tomorrow, man.  I'm confused :-)

maikthomas: @behei-vonage I tested this locally and it's not working for me with your change. I think the `""` needs to go.

Using your version:
```shell
TARGET_BRANCH=main
BASE_BRANCH=rc-mike
echo TARGET_BRANCH $TARGET_BRANCH
echo BASE_BRANCH $BASE_BRANCH
if [[ "$TARGET_BRANCH" == "main" && "$BASE_BRANCH" != "develop" && "$BASE_BRANCH" != "rc-*"  ]]; then
  echo "branch NOT rc or develop"
else
  echo "branch WAS rc or develop"
fi
```
output:
>branch NOT rc or develop

this ❌  fails to detect the *rc-*

Now removing the quotes:
```shell
TARGET_BRANCH=main
BASE_BRANCH=rc-mike
echo TARGET_BRANCH $TARGET_BRANCH
echo BASE_BRANCH $BASE_BRANCH
if [[ "$TARGET_BRANCH" == "main" && "$BASE_BRANCH" != "develop" && "$BASE_BRANCH" != rc-*  ]]; then
  echo "branch NOT rc or develop"
else
  echo "branch WAS rc or develop"
fi
```

Output: 
> branch WAS rc or develop

✅ This correct tested for the *rc-* part

maikthomas (changes requested):

.github/workflows/auto-set-base.yml
maikthomas: @behei-vonage 

>For example, the following:

>BASE_BRANCH="rc-2.19.2"
if [[ "$BASE_BRANCH" != "rc-*" ]]; then
    echo "matches"
else
    echo "no matches"
fi
also returned matches. It's probably just a matter of preference at this point 😉

Please note here that you are logging matches in the if for the NOT equals operator, rather than in the else branch.

behei-vonage: thanks for the explanation Mike I pushed it up with your suggestion 🙏

v-kpheng (approved):
LGTM! 💪 🚀

maikthomas (approved):
LGTM!

dwivedisachin:
Tested LGTM!! 🚀

PR #102, opened 2025-02-27
VIDCS-3397: Pinned video aspect ratio fix

maikthomas:
#### What is this PR doing?
- Fix the aspect ratio for pinned videos when multiple participants are pinned
Before:
image
After:
image

- Change window size with different pinned participants and ensure it looks good
- ensure single pinned / screenshare / active-speaker view still 

Resolves [VIDCS-3397](https://jira.vonage.com/browse/VIDCS-3397)

#### Checklist


frontend/src/hooks/useLayoutManager.tsx
```diff
@@ -4,7 +4,8 @@ import LayoutManager from '../utils/layoutManager';
 
 export type GetLayout = (
   containerDimensions: { height: number; width: number },
-  boxes: Element[]
+  boxes: Element[],
+  fixBigRatio: boolean
```
v-kpheng, line 8: Nit: the name is leaking an implementation detail.  Reading through the code, it looks like a configuration for prioritizing landscape mode over portrait mode.

Maybe this can be renamed to reflect that somehow?

v-kpheng (approved):
LGTM! 💪 🚀

cpettet (dismissed):
LGTM Great job! :shipit:

frontend/src/utils/helpers/getLayoutBoxes/getLayoutBoxes.ts
```diff
@@ -26,15 +26,22 @@ const getLayoutBoxes = ({
   if (!wrapRef.current) {
     return {};
   }
-  const fixBigRatio = layoutProps.pinnedSubscriberCount > 1;
+  // For multiple pinned participants with no screenshare we make all large tiles landscape
+  const fixLargeTilesLandscape =
```
v-kpheng, line 30: Sounds like "desktopView" or "laptopView" to me :-)

maikthomas: hmm, that's not what it is... so this is probably a big name, tbh this is why I avoided trying to make an abstraction because it's not super clear either way.
Another option I guess is to just pass the context into the layout manager rather than the other way around...

frontend/src/hooks/useLayoutManager.tsx
```diff
@@ -1,12 +1,7 @@
 import { useRef } from 'react';
-import { Box, Element } from 'opentok-layout-js';
 import LayoutManager from '../utils/layoutManager';
 
-export type GetLayout = (
-  containerDimensions: { height: number; width: number },
-  boxes: Element[]
-) => Box[];
-
+export type GetLayout = InstanceType<typeof LayoutManager>['getLayout'];
```
maikthomas, line 4: This defines the type as the one used by the class method rather than redefining it here, since changes in the class were not picked up but this type checking. Now any changes in the class will show up where this is used.

v-kpheng (approved):
LGTM! 💪 🚀

cpettet (approved):
LGTM Great job! :shipit:

behei-vonage:
tested, LGTM 🚀

PR #139, opened 2025-03-31
VIDCS-3053: Application does not remember camera/mic settings

behei-vonage:
#### What is this PR doing?

This PR stores the selected audio/video device in local storage so that it can be used once the user joins a new session.

Checkout this branch.
Navigate to a meeting room, select a different audio/video device than you normally use (your default devices).
Close the tab.
Join the meeting room again, notice that it has been initialized with the devices you have previously selected.
Close the tab.
Disconnect the audio/video device that you were using. 
Join the meeting room again, notice that it has been initialized successfully and with the device that is your default one.

Perform similar tests for the waiting room. 

Resolves [VIDCS-3053](https://jira.vonage.com/browse/VIDCS-3053)

#### Checklist
[ ] Resolves a `Known Issue`.


frontend/src/utils/getCurrentDeviceId/getCurrentDeviceId.tsx
```diff
@@ -0,0 +1,41 @@
+let cachedDevices: MediaDeviceInfo[] = [];
```
maikthomas, line 1: Keeping stuff in file storage like this will make it harder to test. Prefer to create stuff in closure and create it via a factory function. Then in the react interface you can use a ref as a holder to ensure its only created once

frontend/src/utils/getCurrentDeviceId/getCurrentDeviceId.tsx
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const refreshDevices = () => {
+  updateDeviceList();
+  if (navigator.mediaDevices) {
+    navigator.mediaDevices.addEventListener('devicechange', updateDeviceList);
+  }
+};
+
+refreshDevices();
```
maikthomas, line 16: likewise with this invocation at a file level

frontend/src/Context/user.tsx
```diff
@@ -49,6 +52,13 @@ const UserProvider = ({ children }: UserProviderProps): ReactElement => {
   const noiseSuppression = window.localStorage.getItem('noiseSuppression') === 'true';
   const blur = window.localStorage.getItem('backgroundBlur') === 'true';
   const name = window.localStorage.getItem('username') ?? '';
+  const deviceManagerRef = useRef<DeviceManagerType | null>(null);
+  if (!deviceManagerRef.current) {
+    deviceManagerRef.current = createDeviceManager();
+  }
+  deviceManagerRef.current?.updateDeviceList();
```
maikthomas, line 59: For the assignation this can be simplified like:
https://github.com/Vonage/vonage-video-react-app/blob/develop/frontend/src/Context/SessionProvider/session.tsx#L154

and If you need to do any extra logic then it should in a hook

behei-vonage: hmm I don't think we can since ActiveSpeakerTracker is a class and the device manager isn't 🤔

frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.spec.tsx
```diff
@@ -69,15 +69,18 @@ describe('usePreviewPublisher', () => {
...
   describe('initLocalPublisher', () => {
-    it('should call initPublisher', () => {
+    it('should call initPublisher', async () => {
       mockedInitPublisher.mockReturnValue(mockPublisher);
       const { result } = renderHook(() => usePreviewPublisher());
+
       result.current.initLocalPublisher();
 
-      expect(mockedInitPublisher).toHaveBeenCalled();
+      await waitFor(() => {
```
behei-vonage, line 78: Note: had to make these `waitFor` since inside `initLocalPublisher` we have an `await` now. Without `waitFor` the spy isn't called

frontend/src/Context/user.tsx
maikthomas: You can still instantiate the ref with createDeviceManager() directly,
`updateDeviceList` should be in a hook, although maybe createDeviceManager should be invoking it if it needs to be called ? 🤔 I need to understand the flow here better

frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.spec.tsx
maikthomas: why not just `await` initLocalPublisher?

frontend/src/utils/localStorageMock.ts
maikthomas, line 1: I'm wondering if this should a testUtils or mocks folder, or both

frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.spec.tsx
behei-vonage: oh yea that works, too. had to change some `act`s to `await` though but yea that works, thanks 🙌

frontend/src/pages/MeetingRoom/MeetingRoom.spec.tsx
```diff
@@ -132,6 +135,7 @@ describe('MeetingRoom', () => {
       getAllMediaDevices: vi.fn(),
       allMediaDevices,
     });
+    (usePublisherOptions as Mock).mockReturnValue({});
```
behei-vonage, line 138: this needed to be mocked due to this: https://github.com/Vonage/vonage-video-react-app/pull/139/files#diff-98d48a076c77e93ec230d6f50b3851675c8168242c0031490ed6a918f4d87f5eR62-R65

frontend/src/Context/PublisherProvider/usePublisherOptions/usePublisherOptions.spec.tsx
```diff
@@ -42,57 +44,90 @@ const mockUserContextWithCustomSettings = {
...
-      audioFallback: {
-        publisher: true,
-      },
-      audioFilter: {
-        type: 'advancedNoiseSuppression',
-      },
-      videoFilter: undefined,
-      name: '',
-      initials: '',
+    await waitFor(() => {
```
behei-vonage, line 73: since `usePublisherOptions` have been moved to a `useEffect`, this needed to be `waitFor`'ed 🙂

frontend/src/Context/PublisherProvider/usePublisherOptions/usePublisherOptions.spec.tsx
```diff
@@ -42,57 +44,90 @@ const mockUserContextWithCustomSettings = {
...
+      });
     });
   });
 
   it('should not have advanced noise suppression if not supported by browser', async () => {
     vi.spyOn(OT, 'hasMediaProcessorSupport').mockReturnValue(false);
     mockUseUserContext.mockImplementation(() => mockUserContextWithDefaultSettings);
     const { result } = renderHook(() => usePublisherOptions());
-    expect(result.current.audioFilter).toBe(undefined);
+    expect(result.current?.audioFilter).toBe(undefined);
```
behei-vonage, line 98: since I changed the `usePublisherOptions` to be potentially `null` - in order to catch the case that the audio source and video source are not set, this needed a `?`

frontend/src/Context/PublisherProvider/usePublisher/usePublisher.tsx
```diff
@@ -76,8 +77,8 @@ const usePublisher = (): PublisherContextType => {
   const [isPublishing, setIsPublishing] = useState(false);
   const publisherOptions = usePublisherOptions();
   const [isForceMuted, setIsForceMuted] = useState<boolean>(false);
-  const [isVideoEnabled, setIsVideoEnabled] = useState(!!publisherOptions.publishVideo);
-  const [isAudioEnabled, setIsAudioEnabled] = useState(!!publisherOptions.publishAudio);
+  const [isVideoEnabled, setIsVideoEnabled] = useState(false);
+  const [isAudioEnabled, setIsAudioEnabled] = useState(false);
```
v-kpheng, line 81: What's the reasoning behind setting -- or correcting, rather -- the state later, as opposed to initializing it with the correct value first?  Just curious...

frontend/src/Context/PublisherProvider/usePublisherOptions/usePublisherOptions.tsx
```diff
@@ -7,49 +7,57 @@ import {
 } from '@vonage/client-sdk-video';
 import useUserContext from '../../../hooks/useUserContext';
 import getInitials from '../../../utils/getInitials';
+import DeviceManager from '../../../utils/DeviceManager';
 
 /**
  * React hook to get PublisherProperties combining default options and options set in UserContext
- * @returns {PublisherProperties} publisher properties object
+ * @returns {PublisherProperties | null} publisher properties object
```
v-kpheng, line 14: What's the benefit of making this nullable, as opposed to having the caller use an empty object?

frontend/src/utils/DeviceManager/deviceManager.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Device manager that retrieves the stored device ID for a given device type (audio or video)
```
v-kpheng, line 2: Pedantic: a device manager allows users to manage a device, like adding, removing, or editing it.  This seems more like a `DeviceStore`.

I'll shut up now 🙊

frontend/src/utils/DeviceManager/deviceManager.ts
```diff
@@ -0,0 +1,20 @@
...
+  async init() {
+    if (navigator.mediaDevices) {
+      this.devices = await navigator.mediaDevices.enumerateDevices();
+    }
+  }
+
+  getConnectedDeviceId(kind: 'audioinput' | 'videoinput'): string | undefined {
+    const key = kind === 'videoinput' ? 'videoSource' : 'audioSource';
+    const storedId = window.localStorage.getItem(key);
+    return this.devices.find((device) => device.kind === kind && device.deviceId === storedId)
```
v-kpheng, line 17: Do we need this filter?  Naively, seems like we can just store the `deviceId` in `localStorage` as well.

frontend/src/Context/PublisherProvider/usePublisherOptions/usePublisherOptions.spec.tsx
v-kpheng: Does it need to be nullable?  Isn't an empty object sufficient?

frontend/src/Context/PublisherProvider/usePublisherOptions/usePublisherOptions.tsx
```diff
@@ -7,49 +7,57 @@ import {
...
+    const setOptions = async () => {
+      if (!deviceManagerRef.current) {
+        deviceManagerRef.current = new DeviceManager();
+        await deviceManagerRef.current.init();
+      }
+
+      const videoSource = deviceManagerRef.current.getConnectedDeviceId('videoinput');
+      const audioSource = deviceManagerRef.current.getConnectedDeviceId('audioinput');
+
+      const { name, noiseSuppression, blur, publishAudio, publishVideo } = user.defaultSettings;
```
v-kpheng, line 32: IMO, it's odd that we have these settings in `UserContext` but we don't have the other settings above in it.  Instead, they're in `DeviceManager`.

If the media sources don't belong in `UserContext`, then maybe we can add them in `AppContext`, `SystemContext`, `ConfigContext` or something.

v-kpheng (commented):
Looks good! 💪 

That said, left some questions/comments.  Also, didn't review the test code yet.

frontend/src/utils/DeviceManager/deviceManager.ts
behei-vonage: Yes, this is meant to check if the deviceId that was previously stored in the `localStorage` is still connected to the device and has not been unplugged 🙏

frontend/src/Context/PublisherProvider/usePublisher/usePublisher.tsx
behei-vonage: ah the reason is that I changed the publisher options to potentially be null so it was not liking that

frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.tsx
```diff
@@ -119,6 +122,7 @@ const usePreviewPublisher = (): PreviewPublisherContextType => {
       }
       publisherRef.current.setAudioSource(deviceId);
       setLocalAudioSource(deviceId);
+      window.localStorage.setItem('audioSource', deviceId);
```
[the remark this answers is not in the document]
behei-vonage: I think I've attempted that before but it didn't go so well 😅

frontend/src/utils/DeviceManager/deviceManager.ts
behei-vonage: sure, actually that's a great suggestion! I was struggling to pick the right name for this and `DeviceManager` sounded good to me but I like `DeviceStore` better

frontend/src/Context/PublisherProvider/usePublisherOptions/usePublisherOptions.spec.tsx
behei-vonage: I think it does in this case because of using this `useEffect` [here](https://github.com/Vonage/vonage-video-react-app/pull/139/files#diff-2bf3c35c2aa956b3ba82d7b5bea5842607a6d703df923b5ecc4ec958b427d261R101-R108) and [here](https://github.com/Vonage/vonage-video-react-app/pull/139/files#diff-ed35da1d82a0249b96617d09ffc6eb1e33f80541beed2e7292d0371466b2fac6R22-R27).

if I do `publisherOptions` as an empty object instead of null doing `if (!publisherOptions)` won't work and doing stuff like `if (Object.keys(publisherOptions).length === 0)` isn't pretty

frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.tsx
```diff
@@ -228,10 +235,19 @@ const usePreviewPublisher = (): PreviewPublisherContextType => {
           }
         : undefined;
 
+    if (!isDeviceManagerReady) {
+      await deviceManagerRef.current.init();
```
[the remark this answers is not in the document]
behei-vonage: I don't think so. worst case scenario is that `enumerateDevices` returns an empty array of devices. in that case, everything is going to be defined with an `undefined` which is accepted by the sdk

frontend/src/Context/PublisherProvider/usePublisherOptions/usePublisherOptions.tsx
behei-vonage: would that refactor be a blocker for you or could I do in a follow up PR?

frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.tsx
```diff
@@ -228,10 +235,19 @@ const usePreviewPublisher = (): PreviewPublisherContextType => {
           }
         : undefined;
 
+    if (!isDeviceStoreReady) {
+      await deviceStoreRef.current.init();
+      videoSource = deviceStoreRef.current.getConnectedDeviceId('videoinput');
+      audioSource = deviceStoreRef.current.getConnectedDeviceId('audioinput');
+      setIsDeviceStoreReady(true);
```
maikthomas, line 242: I think you could probably extract all of this stuff away into a `useStoredDevices()` (or similar) hook  that handles this and just returns a couple variables of state in an object like maybe:
```js
const {deviceStoreReady, storedAudioDevice, storedVideoDevice} = useStoredDevices();
```

frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.tsx
```diff
@@ -228,10 +235,19 @@ const usePreviewPublisher = (): PreviewPublisherContextType => {
           }
         : undefined;
 
+    if (!isDeviceStoreReady) {
```
maikthomas, line 238: Is there actually a scenario where this is true when called here if init hasn't been called?
or are we invoking this function twice? I'm a bit lost on the flow of this function

frontend/src/Context/PublisherProvider/usePublisher/usePublisher.tsx
```diff
@@ -162,31 +172,35 @@ const usePublisher = (): PublisherContextType => {
...
-      // Add listeners synchronously as some events could be fired before callback is invoked
-      addPublisherListeners(publisher);
-      publisherRef.current = publisher;
-    } catch (error) {
-      if (error instanceof Error) {
-        console.warn(error.stack);
+  const initializeLocalPublisher = useCallback(
+    (options?: PublisherProperties) => {
+      try {
+        const publisher = initPublisher(undefined, options);
```
maikthomas, line 192: if options is undefined do we want to be doing this?

frontend/src/components/MeetingRoom/InputDevices/InputDevices.tsx
```diff
@@ -38,6 +38,7 @@ const InputDevices = ({ handleToggle, customLightBlueColor }: InputDevicesProps)
     })?.deviceId;
     if (audioDeviceId) {
       publisher?.setAudioSource(audioDeviceId);
+      window.localStorage.setItem('audioSource', audioDeviceId);
```
maikthomas, line 41: We might want to do one of both of two things:
- extract this interaction with localStorage to its own file to track where we interact with localstorage
- use a shared constant for the keys
currently there will be no error if this line was accidentally 
```ts
      window.localStorage.setItem('audioDevice', audioDeviceId);
```

frontend/src/Context/PublisherProvider/usePublisher/usePublisher.tsx
```diff
@@ -32,7 +33,7 @@ export type AccessDeniedEvent = Event<'accessDenied', Publisher> & {
 };
 
 export type PublisherContextType = {
-  initializeLocalPublisher: () => void;
+  initializeLocalPublisher: (options?: PublisherProperties) => void;
```
maikthomas, line 36: Hmm, I understand that publisherProperties from the hook can be null while they're 'loading', but this isn't an optional parameter is it? 
In fact you do an early return if options isn't defined where this is invoked

frontend/src/utils/SpeakingDetector/speakingDetector.ts
```diff
@@ -89,7 +93,6 @@ class SpeakingDetector extends EventEmitter {
 
       this.source.connect(this.analyser);
 
-      this.analyser.fftSize = 2048;
```
maikthomas, line 92: 🤔 how come?

frontend/src/utils/SpeakingDetector/speakingDetector.ts
```diff
@@ -77,7 +77,11 @@ class SpeakingDetector extends EventEmitter {
    */
   turnSpeakingDetectorOn = async () => {
     try {
+      if (!this.selectedMicrophoneId) {
+        return;
+      }
```
maikthomas, line 82: how come?

frontend/src/utils/DeviceStore/deviceStore.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Device store that retrieves the stored device ID for a given device type (audio or video)
+ * and checks if it is still connected.
+ */
+export default class DeviceStore {
+  private devices: MediaDeviceInfo[] = [];
+
+  async init() {
+    if (navigator.mediaDevices) {
+      this.devices = await navigator.mediaDevices.enumerateDevices();
```
maikthomas, line 10: I'm assuming this code flow still works  ok when we don't have user permissions yet

maikthomas (commented):
A few questions to understand a couple things better

frontend/src/utils/SpeakingDetector/speakingDetector.ts
behei-vonage: this actually needs to be put back. I think I accidentally deleted it. apologies

frontend/src/utils/SpeakingDetector/speakingDetector.ts
behei-vonage: I started after the changes I've done that the speaking detector was returning undefined. the useSpeakingIndicator hook gets called a few times so it gets initialized okay but adding this `if` statement removes a console log error

frontend/src/Context/PublisherProvider/usePublisher/usePublisher.tsx
behei-vonage: changed the options to be required and not optional so going to resolve this one and I think another comment around this 🙏

frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.tsx
behei-vonage: hmm if that's okay I'd like to keep it as-is, at least for now. I tried moving it all in its own hook but I'm not able to get it to work 😅 I tried variety of things but yea no luck

frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.tsx
```diff
@@ -69,6 +70,8 @@ const usePreviewPublisher = (): PreviewPublisherContextType => {
   const [isAudioEnabled, setIsAudioEnabled] = useState(true);
   const [localVideoSource, setLocalVideoSource] = useState<string | undefined>(undefined);
   const [localAudioSource, setLocalAudioSource] = useState<string | undefined>(undefined);
+  const deviceStoreRef = useRef<DeviceStore>(new DeviceStore());
+  const [isDeviceStoreReady, setIsDeviceStoreReady] = useState<boolean>(false);
```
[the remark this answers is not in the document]
behei-vonage: hmm maybe it's not needed - it seems to work just fine without it. I _think_ at first I was using the `init` method of the `ref` outside of the `useCallback` so maybe it was needed there

frontend/src/utils/storage.ts
```diff
@@ -0,0 +1,12 @@
+export const STORAGE_KEYS = {
+  AUDIO_SOURCE: 'audioSource',
+  VIDEO_SOURCE: 'videoSource',
+};
```
behei-vonage, line 4: Note: we use `setItem` and `getItem` for a few other things such as background blur etc but I didn't want to include changes to that in this PR since it seems unrelated. I'll create a follow up ticket and change it there

frontend/src/Context/PublisherProvider/usePublisher/usePublisher.spec.tsx
```diff
@@ -83,7 +83,7 @@ describe('usePublisher', () => {
 
       const { result } = renderHook(() => usePublisher());
       act(() => {
-        result.current.initializeLocalPublisher();
+        result.current.initializeLocalPublisher({});
```
behei-vonage, line 86: Note: all of these were needed since I made `initializeLocalPublisher`'s parameter of `options` to be required instead of optional

frontend/src/Context/PublisherProvider/usePublisher/usePublisher.tsx
```diff
@@ -76,8 +77,12 @@ const usePublisher = (): PublisherContextType => {
...
   const publisherOptions = usePublisherOptions();
   const [isForceMuted, setIsForceMuted] = useState<boolean>(false);
-  const [isVideoEnabled, setIsVideoEnabled] = useState(!!publisherOptions.publishVideo);
-  const [isAudioEnabled, setIsAudioEnabled] = useState(!!publisherOptions.publishAudio);
+  const [isVideoEnabled, setIsVideoEnabled] = useState<boolean>(
+    publisherOptions?.publishVideo ?? false
+  );
+  const [isAudioEnabled, setIsAudioEnabled] = useState<boolean>(
+    publisherOptions?.publishAudio ?? false
+  );
```
behei-vonage, line 85: @maikthomas, I followed your suggestion here but I still need to do the `useEffect` below. otherwise the publisher in the meeting room initialized with `false` and not what's in the `publisherOptions`

maikthomas: mmm, in that case is this doing anything? or is `publisherOptions` always undefined in this scenario?

frontend/src/utils/DeviceStore/deviceStore.ts
```diff
@@ -0,0 +1,20 @@
...
+    }
+  }
+
+  getConnectedDeviceId(kind: 'audioinput' | 'videoinput'): string | undefined {
+    const key = kind === 'videoinput' ? 'videoSource' : 'audioSource';
+    const storedId = window.localStorage.getItem(key);
```
maikthomas, line 16: This can be updated to use your storage methods and keys

maikthomas (approved):
Nice one, LGTM, just a couple small comments

frontend/src/Context/PublisherProvider/usePublisher/usePublisher.tsx
behei-vonage: I think `publisherOptions` is always `undefined` when we set the initial state of this. that's why I was setting it to `false` originally

frontend/src/utils/DeviceStore/deviceStore.ts
behei-vonage: oh yes, definitely, I forgot to do that 😅

frontend/src/utils/DeviceStore/deviceStore.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+      { deviceId: 'device-123', kind: 'videoinput' } as MediaDeviceInfo,
+    ]);
+
+    await deviceStore.init();
+    const result = deviceStore.getConnectedDeviceId('videoinput');
+
+    expect(result).toBe('device-123');
+  });
+
+  test('returns undefined if stored device is not connected', async () => {
```
v-kpheng, line 45: Nice 💪

v-kpheng (approved):
LGTM! 💪 🚀 

I have some nits, but I'll shut up 🙊

DeliaTok:
looks good. verified on Chrome, Safari and FF.

PR #161, opened 2025-05-07
VIDCS-3699: Implement server-side controls of captions 

behei-vonage:
#### What is this PR doing?

This PR implements server-side controls of captions. 

This needs to be tested with both `vonage` and `opentok` as `VIDEO_SERVICE_PROVIDER`s in your `.env` file in the backend. For that, you'll need to have both Vonage applicationId and OpenTok API key.

To test this, you will also need to have Postman installed. 
Once that is done, run the app by doing `yarn dev`.
In Postman, make a `GET` request to `localhost:3345/session/<room_name_of_your_choice>` (notice no trailing slash) - this is going to create a room for you - captions cannot be enabled unless a room has already been created. 
After that, make a `POST` request to `localhost:3345/session/<room_name_of_your_choice>/enableCaptions`.
Notice that you are getting a return that looks something like this:

```
{
    "captions": {
        "captionsId": "3f9ed25c-a12c-4ff4-abe9-428f4371dae2"
    },
    "status": 200
}
```
Copy-paste the `captionsId` you got back and make a `POST` request to `localhost:3345/session/<room_name_of_your_choice>/<captionsId_here>/disableCaptions`. 
If using the example above, `<captionsId_here>` would be `3f9ed25c-a12c-4ff4-abe9-428f4371dae2`. 

Resolves [VIDCS-3699](https://jira.vonage.com/browse/VIDCS-3699)

#### Checklist
[ ] Resolves a `Known Issue`.

### DIFF
--- a/backend/routes/session.ts
+++ b/backend/routes/session.ts
@@ -91,15 +91,11 @@ sessionRouter.post(
       const { room: roomName } = req.params;
       const sessionId = await sessionService.getSession(roomName);
       if (sessionId) {
-        if (videoService.enableCaptions) {
-          const captions = await videoService.enableCaptions(sessionId);
-          res.json({
-            captions,
-            status: 200,
-          });
-        } else {
-          res.status(500).json({ message: 'Captions service is unavailable' });
-        }
+        const captions = await videoService.enableCaptions(sessionId);
+        res.json({
+          captions,
+          status: 200,
+        });
       } else {
         res.status(404).json({ message: 'Room not found' });
       }
@@ -115,17 +111,11 @@ sessionRouter.post(
   async (req: Request<{ room: string; captionId: string }>, res: Response) => {
     try {
       const { captionId } = req.params;
-      if (captionId) {
-        if (videoService.disableCaptions) {
-          const responseCaptionId = await videoService.disableCaptions(captionId);
-          res.json({
-            captionId: responseCaptionId,
-            status: 200,
-          });
-        } else {
-          res.status(500).json({ message: 'Captions service is unavailable' });
-        }
-      }
+      const responseCaptionId = await videoService.disableCaptions(captionId);
+      res.json({
+        captionId: responseCaptionId,
+        status: 200,
+      });
     } catch (error: unknown) {
       res.status(500).send({ message: (error as Error).message ?? error });
     }
--- a/backend/routes/session.ts
+++ b/backend/routes/session.ts
@@ -84,4 +84,52 @@ sessionRouter.get('/:room/archives', async (req: Request<{ room: string }>, res:
   }
 });
 
+sessionRouter.post(
+  '/:room/enableCaptions',
+  async (req: Request<{ room: string }>, res: Response) => {
+    try {
+      const { room: roomName } = req.params;
+      const sessionId = await sessionService.getSession(roomName);
+      if (sessionId) {
+        if (videoService.enableCaptions) {
+          const captions = await videoService.enableCaptions(sessionId);
+          res.json({
+            captions,
+            status: 200,
+          });
+        } else {
+          res.status(500).json({ message: 'Captions service is unavailable' });
+        }
+      } else {
+        res.status(404).json({ message: 'Room not found' });
+      }
+    } catch (error: unknown) {
+      const errorMessage = error instanceof Error ? error.message : 'Unknown error occurred';
+      res.status(500).json({ message: errorMessage });
+    }
+  }
+);
+
+sessionRouter.post(
+  '/:room/:captionId/disableCaptions',
+  async (req: Request<{ room: string; captionId: string }>, res: Response) => {
+    try {
+      const { captionId } = req.params;
+      if (captionId) {
+        if (videoService.disableCaptions) {
+          const responseCaptionId = await videoService.disableCaptions(captionId);
+          res.json({
+            captionId: responseCaptionId,
+            status: 200,
+          });
+        } else {
+          res.status(500).json({ message: 'Captions service is unavailable' });
+        }
+      }
+    } catch (error: unknown) {
+      res.status(500).send({ message: (error as Error).message ?? error });
+    }
+  }
+);
+
 export default sessionRouter;
--- a/backend/types/opentok-jwt-d.ts
+++ b/backend/types/opentok-jwt-d.ts
@@ -1,0 +1,4 @@
+declare module 'opentok-jwt' {
+  // eslint-disable-next-line import/prefer-default-export
+  export function projectToken(apiKey: string, apiSecret: string, expire?: number): string;
+}
--- a/backend/videoService/opentokVideoService.ts
+++ b/backend/videoService/opentokVideoService.ts
@@ -106,37 +106,36 @@ class OpenTokVideoService implements VideoService {
     const projectJWT = projectToken(this.config.apiKey, this.config.apiSecret, expires);
     const captionURL = `${this.API_URL}/${this.config.apiKey}/captions`;
 
-    const requestToken = this.generateToken(sessionId);
-    const { token } = requestToken;
+    const { token } = this.generateToken(sessionId);
     const captionAxiosPostBody = {
       sessionId,
       token,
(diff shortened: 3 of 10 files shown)


backend/types/opentok-jwt-d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'opentok-jwt' {
```
behei-vonage, line 1: Note: this was needed since I'm using the `opentok-jwt` package to generate a project token; however, the package is not TypeScript friendly so this was needed. I also cannot it export with a default; therefore, disabling lint since it needs to be imported like this:
```
import { projectToken } from 'opentok-jwt`
```
therefore not supporting the default export approach

backend/videoService/tests/opentokVideoService.test.ts
```diff
@@ -0,0 +1,118 @@
+import { describe, expect, it, jest } from '@jest/globals';
```
behei-vonage, line 1: Note: if you notice some differences between this test file and `vonageVideoService.test.ts`, no you don't 😆 
just kidding - there are some slight difference since the `opentok` package uses callback and exports itself without a `default` but with `module.exports = OpenTok`.

also you'll notice that in this file I'm also mocking `axios` - that is because we have to use `axios` in `opentokVideoService` but not `vonageVideoService` since the `opentok` library does not support `enableCaptions` and `disableCaptions` methods so we have to do direct REST API requests. I've left some comments about it in the codebase, too.

v-kpheng: Well, to be fair, for tests we prefer WET over DRY ;-)

v-kpheng (approved):
LGTM! 💪 🚀 

Some nice mocking here :-)

backend/routes/session.ts
```diff
@@ -84,4 +84,42 @@ sessionRouter.get('/:room/archives', async (req: Request<{ room: string }>, res:
...
+      const { room: roomName } = req.params;
+      const sessionId = await sessionService.getSession(roomName);
+      if (sessionId) {
+        const captions = await videoService.enableCaptions(sessionId);
+        res.json({
+          captions,
+          status: 200,
+        });
+      } else {
+        res.status(404).json({ message: 'Room not found' });
```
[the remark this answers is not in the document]
behei-vonage: perhaps but I'd like to keep it as-is to be consistent with the other routes we have. 
though, let me know if I'm wrong - 
if this line throws: 
```
const captions = await videoService.enableCaptions(sessionId);
```
then it will be caught by:
```
catch (error: unknown) {
      const errorMessage = error instanceof Error ? error.message : 'Unknown error occurred';
      res.status(500).json({ message: errorMessage });
    }
```
however, if it so happens that `sessionId` is undefined, instead of making the `enableCaptions` request, it will go straight to: 
```
 res.status(404).json({ message: 'Room not found' });
```
which seems to be covering all of our scenarios, right?

backend/videoService/vonageVideoService.ts
```diff
@@ -59,6 +66,22 @@ class VonageVideoService implements VideoService {
     await this.vonageVideo.stopArchive(archiveId);
     return 'Archive stopped successfully';
   }
+
+  async enableCaptions(sessionId: string): Promise<EnableCaptionResponse> {
+    const requestToken = this.generateToken(sessionId);
+    const { token } = requestToken;
+    return this.vonageVideo.enableCaptions(sessionId, token);
```
cpettet, line 73: Our sample node repo has this wrapped in a `try-catch`. Can we ensure there's proper error-handling from the vonage video service?

cpettet, line 73: Should we specify the `languageCode` and `partialCaptions`, too?

backend/videoService/videoServiceInterface.ts
```diff
@@ -7,4 +7,6 @@ export interface VideoService {
   startArchive(roomName: string, sessionId: string): Promise<Archive | SingleArchiveResponse>;
   stopArchive(archiveId: string): Promise<string>;
   listArchives(sessionId: string): Promise<Archive[] | SingleArchiveResponse[] | undefined>;
+  enableCaptions(sessionId: string): Promise<EnableCaptionResponse | string | undefined>;
```
cpettet, line 10: When do we return `string` or `undefined`?

cpettet (commented):
Looks good so far! Just have a few comments/questions. Let me know what you think!

backend/videoService/videoServiceInterface.ts
behei-vonage: great catch, we only return `EnableCaptionResponse`. The other two came from before I added the said `EnableCaptionResponse`. I'll remove it. thanks! 💪

backend/videoService/vonageVideoService.ts
behei-vonage: I was mostly going for consistency with the start/stop archive here but definitely, having the `try-catch` has its benefit. I'm also going to add it to the opentok video service for consistency. thanks Chris!

backend/videoService/vonageVideoService.ts
```diff
@@ -59,6 +67,37 @@ class VonageVideoService implements VideoService {
     await this.vonageVideo.stopArchive(archiveId);
     return 'Archive stopped successfully';
   }
+
+  async enableCaptions(sessionId: string): Promise<EnableCaptionResponse> {
+    const requestToken = this.generateToken(sessionId);
+    const { token } = requestToken;
+
+    try {
+      const captionOptions: CaptionOptions = {
```
behei-vonage, line 76: Note: if you notice that `languageCode` is missing here compared to `opentokVideoService`, no you didn't 
when I was adding the caption options,  I found an issue with `vonage-node-sdk` and I filed a PR to fix it here: https://github.com/Vonage/vonage-node-sdk/pull/993
the big issue is that it accepts `languageCode` as `en-us` and not `en-US`; therefore, making it invalid and I'm not able to successfully enable captions.
for the time being, I decided to skip adding the `languageCode` as `en-US` is the default option anyway. once the DevRel team merges in / deploys the new version that has the change I added, I'll add it back here for consistency with `opentokVideoService`

cpettet (approved):
LGTM Great job! :shipit:

backend/videoService/opentokVideoService.ts
```diff
@@ -106,37 +106,47 @@ class OpenTokVideoService implements VideoService {
...
     const captionURL = `${this.API_URL}/${this.config.apiKey}/captions`;
 
-    const requestToken = this.generateToken(sessionId);
-    const { token } = requestToken;
-    const captionAxiosPostBody = {
-      sessionId,
-      token,
+    const { token } = this.generateToken(sessionId);
+    const captionOptions = {
+      // The following language codes are supported: en-US, en-AU, en-GB, fr-FR, fr-CA, de-DE, hi-IN, it-IT, pt-BR, ja-JP, ko-KR, zh-CN, zh-TW
```
v-kpheng, line 111: Nit: maybe link to where this is documented instead?

backend/videoService/opentokVideoService.ts
```diff
@@ -106,37 +106,47 @@ class OpenTokVideoService implements VideoService {
...
       languageCode: 'en-US',
+      // The maximum duration of the captions in seconds. The default is 14,400 seconds (4 hours).
```
v-kpheng, line 113: Same nit above applies here too.

backend/videoService/opentokVideoService.ts
```diff
@@ -106,37 +106,47 @@ class OpenTokVideoService implements VideoService {
...
       maxDuration: 1800,
+      // Enabling partial captions allows for more frequent updates to the captions.
+      // This is useful for real-time applications where the captions need to be updated frequently.
+      // However, it may also increase the number of inaccuracies in the captions.
```
v-kpheng, line 117: Excellent comment! 💪

v-kpheng (approved):
LGTM! 💪 🚀

DeliaTok:
looks good!

PR #183, opened 2025-06-30
VIDCS-3919: Auto Dependencies Update job failing

behei-vonage:
#### What is this PR doing?

This PR fixes the "auto-update-dependencies" job. As a reminder, this one gets run after PRs are merged into `develop`. After latest release, I noticed that the job wasn't running. 

This broke after `axios` was added as a part of `resolutions` in `package.json` so we had some version conflicts. Since it was added there, though, we no longer need to manually install it; only need to run `yarn install` to make sure that all the dependencies are there. 

Notice that the job has successfully ran and committed changes [here](https://github.com/Vonage/vonage-video-react-app/pull/183/commits/268ea48cb053e874c89e0f7fa547f6b074ca6305). 

Resolves [VIDCS-3919](https://jira.vonage.com/browse/VIDCS-3919)

#### Checklist
[ ] Resolves a `Known Issue`.


.github/workflows/auto-update-dependencies.yml
```diff
@@ -9,7 +9,7 @@ permissions:
   contents: read
 
 env:
-  BRANCH_NAME: ${{ github.head_ref }}
+  BRANCH_NAME: ${{ github.ref_name }}
```
behei-vonage, line 12: Note: the value of `head_ref` is an empty string when it's accessed via a direct push to the branch. It is only accessible when doing it via a pull request which is why a previous PR related to this functionality passed the testing. I don't think this was ever actually posting any commits to `develop`, though.
Setting the value to `ref_name` instead will allow us to access that `develop` value which is what we want.

v-kpheng (approved):
LGTM! 💪 🚀

cpettet (approved):
LGTM Great job! :shipit:

PR #251, opened 2025-10-30
VIDSOL-281: refactor app config context for scoped reactivity and stable rests

johnny-quesada-developer:
#### What is this PR doing?
This PR refactors the AppConfigContext to isolate updates and stop global re-renders.
It forces components to subscribe only to the specific configuration keys they depend on, eliminating unnecessary re-renders and improving overall performance.
It also introduces the isAppConfigLoaded flag to stabilize initialization and prevent UI flickers.

Additionally, it removes excessive mocking from our test suite to improve reliability and ensure tests reflect real behavior.
This change lays the groundwork for future low-code/no-code configuration flows, where the AppConfigContext will become more dynamic and require granular state updates.

Resolves [VIDSOL-281](https://jira.vonage.com/browse/VIDSOL-281)

#### Checklist
[ ] Resolves a `Known Issue`.


frontend/src/components/HiddenParticipantsTile/HiddenParticipantsTile.spec.tsx
```diff
@@ -77,50 +53,85 @@ describe('HiddenParticipantsTile', () => {
...
+    const button = getByTestId('hidden-participants');
     expect(button).toBeInTheDocument();
     await userEvent.click(button);
 
-    expect(mockToggleParticipantList).not.toHaveBeenCalled();
+    expect(toggleParticipantListSpy).not.toHaveBeenCalled();
   });
 });
+
+function render(ui: ReactElement, options?: SessionProviderWrapperOptions) {
```
behei-vonage: could you please help explain if there is any particular reason (besides personal preference) to use the `function` notation instead of the arrow function notation we've been using throughout the project? I'd prefer to stick to the arrow notation please as to remain consistent; otherwise, we'll have to go in and change everything

johnny-quesada-developer: Hi of course @behei-vonage ,

Actually, I didn’t use the function notation out of personal preference, but because of hoisting. I’m declaring the function as a utility at the end of the file, but I want it to be available everywhere within the file. With an arrow function, I can’t guarantee that it will always be available, but with a regular function declaration, I can. Using the function keyword is more convenient in this case because I don’t even need to worry about where the render call is or anything like that.

With function key:
image

With arrow function:
image

OscarFava (dismissed):
LGTM!

VZaphod (dismissed):
LGTM

rserebrennykov (dismissed):
LGTM

PR #300, opened 2025-12-12
VIDSOL-303: refactor test to use real providers instead of context mocks

Hossein-Movahed:
Refactor test to use real providers instead of context mocks includes explanatory comments

#### What is this PR doing?
- Remove `publisherContext` and `sessionContext` mocks from `MeetingRoom.spec`, 
- Remove `useAudioOutput` context mock form `DeviceSettingsMenu.spec` and `useAudioOutput.spec`,
- Remove context mock from `useBackgroundPublisher.spec`
- Remove context mock from `usePreviewPublisher.spec`
- Add provider wrappers supporting inline per-test configuration  
- Tests validate real data flow through contexts
- Add docs/TESTING.md with full testing philosophy and patterns
- Link testing guide from README.md

Tests passing: 
- `MeetingRoom.spec`: 11 passing tests,(2 removed from code)
- `DeviceSettingsMenu.spec`: 10 passing tests
- `useAudioOutput.spec`: 4 passing tests
- `useBackgroundPublisher.spec`: 10 passing tests
- `usePreviewPublisher.spec`: 10 passing tests

Once we agree on the approach, we can refactor all the tests in batches of 5 or 10 files.

#### Removed tests note:
The 2 removed tests in MeetingRoom.spec rely on quality values from the OpenTok SDK, which cannot be set via initial values or direct mutation. This is explained here instead of keeping them in code.

Resolves [VIDSOL-303](https://jira.vonage.com/browse/VIDSOL-303)

#### Checklist
[ ] Resolves a `Known Issue`.


frontend/src/Context/SessionProvider/session.spec.tsx
```diff
@@ -405,17 +405,16 @@ describe('SessionProvider', () => {
...
-          reconnections: 0,
-        },
-      } as UserType,
-    },
-  });
+type RenderOverrides = {
+  session?: Parameters<typeof makeSessionProviderWrapper>[0];
+};
+
+function render(ui: ReactElement, overrides: RenderOverrides = {}) {
```
johnny-quesada-developer, line 475: What's the reason for changing this `.spec` file? We're not adding new coverage or fixing anything.

On the contrary, I see that removing the `user` value has `zero` effect, which means we're not testing any use case where the user is expected to be present.

Can you please do the following?

1 - Modify **makeSessionProviderWrapper** so it follows the same format as the other wrappers:

```tsx
// Instead of this:
export type SessionProviderWrapperOptions = GenericWrapperOptions<
  typeof SessionProvider,
  typeof SessionContext
> & {
  appConfigOptions?: AppConfigProviderWrapperOptions;
  userOptions?: UserProviderWrapperOptions;
};

// The parameter should look like:
export type SessionProviderWrapperOptions = {
  sessionOptions?: GenericWrapperOptions<typeof SessionProvider, typeof SessionContext>;
  appConfigOptions?: AppConfigProviderWrapperOptions;
  userOptions?: UserProviderWrapperOptions;
};

// So that we can have 100% control of the app state while testing:
function render(
  ui: ReactElement,
  { sessionOptions, userOptions, appConfigOptions }: SessionProviderWrapperOptions = {}
) {
  const { SessionProviderWrapper, ...props } = makeSessionProviderWrapper({
    sessionOptions,
    userOptions,
    appConfigOptions,
  });

  return {
    ...props,
    ...renderBase(ui, { wrapper: SessionProviderWrapper }),
  };
}
```

2 - Then, let's add a test for the only use case inside the session provider that actually requires a user. This will prevent us from losing that functionality during the upcoming migration.

Example:

```tsx
it('should correctly increase reconnection counter', () => {
  // Context should be created per test, otherwise it will break isolation.
  const { userContext, sessionContext } = render(<TestComponent />, {
    userOptions: {
      value: {
        issues: { reconnections: 0 },
      } as UserType,
    },
  });

  expect(userContext.current?.user.issues.reconnections).toBe(0);

  sessionContext.current.vonageVideoClient?.emit('sessionReconnecting');

  expect(userContext.current?.user.issues.reconnections).toBe(1);
});
```

johnny-quesada-developer: Let's also update **makeUserProviderWrapper**.

```tsx
// The parameter should go from:
export type UserProviderWrapperOptions = GenericWrapperOptions<
  typeof UserProvider,
  typeof UserContext
>;

// To:
export type UserProviderWrapperOptions = {
  userOptions: GenericWrapperOptions<typeof UserProvider, typeof UserContext>;
};
```

We should keep this format for all wrappers so that, if we later add new dependencies to the provider, we can easily extend it.

frontend/src/components/MeetingRoom/AudioIndicator/AudioIndicator.spec.tsx
```diff
@@ -28,19 +25,31 @@ describe('AudioIndicator', () => {
     audioLevel: undefined,
   };
 
-  beforeEach(() => {
-    (useSessionContext as Mock).mockReturnValue({ forceMute: mockForceMute });
-    vi.clearAllMocks();
+  // Only include SessionContext - explicitly exclude other providers
+  // to avoid unnecessary dependencies (mediaDevices, Vonage SDK initialization, etc.)
+  const TestWrapper = createTestProviderStack({
```
johnny-quesada-developer: Please do not declare global providers for tests. This is an anti-pattern in unit testing. By doing so, we are causing subsequent tests to depend on the state produced by previous ones. As a result, adding, removing, or modifying a single test can unintentionally break others. This approach is not sustainable. Even running a single test with `.only` for debugging purposes can lead to state leakage and unexpected failures.

Tests must remain isolated. This is precisely why we rely on context instead of global state. Context gives us fine-grained control and proper isolation per test case.

Let's also remove the helper `createTestProviderStack`:

- The helper is not granular. It exposes more providers than necessary, and its default behavior enables all providers. This encourages developers to just `createTestProviderStack()` without configuration, with unnecessary providers being mounted, heavier tests, and increased pipeline execution time.
- The helper does not provide access to the underlying context instances. This makes it impossible to inspect context state or spy on context-level functions.
- The helper does not allow overriding default values. How can I modify the audio output state to reproduce a specific scenario.
- `createTestProviderStack` is less declarative than the approach we were already implementing.

```tsx
// Which providers does this include?
const Wrapper = createTestProviderStack();

// Versus this approach, where we explicitly create a session provider that includes
// only the dependencies required by the SessionProvider, while also exposing access
// to the related contexts.
const {
  SessionProviderWrapper,
  sessionContext,
  userContext,
  appConfigContext,
} = makeSessionProviderWrapper();
```

@OscarFava ?

frontend/src/components/MeetingRoom/AudioIndicator/AudioIndicator.spec.tsx
```diff
@@ -28,19 +25,31 @@ describe('AudioIndicator', () => {
...
-    render(<AudioIndicator {...defaultProps} />);
+    render(<AudioIndicator {...defaultProps} />, { wrapper: TestWrapper });
     const micIcon = screen.getByTestId('MicIcon');
     expect(micIcon).toBeInTheDocument();
   });
 
+  // We can pass providers inline to avoid initializing unused contexts.
   it('renders Mic off icon when participant is muted', () => {
-    render(<AudioIndicator {...defaultProps} hasAudio={false} />);
+    render(<AudioIndicator {...defaultProps} hasAudio={false} />, {
```
johnny-quesada-developer: If I remove this wrapper, there is `zero` effect. The tests keep running...

What we should get from this task is an overall improvement in test reliability. Right now the `SessionContext` has a non-null default value. As a result, components can render outside `<SessionProvider>` boundaries without anyone noticing. This weakens the suite because tests do not fail when the provider is missing.

We should fix this by removing the default context object and making the context nullable. If doing this properly expands the PR scope too much, we can split it into a follow-up task, but we should not ignore this finding.

Proposed change:

```tsx
// BAD: provides a fully populated default object, allowing usage outside the provider
export const SessionContext = createContext<SessionContextType>({
  vonageVideoClient: null,
  disconnect: null,
  joinRoom: null,
  forceMute: null,
  connected: null,
  unreadCount: 0,
  messages: [],
  sendChatMessage: () => {},
  reconnecting: null,
  subscriberWrappers: [],
  activeSpeakerId: undefined,
  layoutMode: 'grid',
  setLayoutMode: () => {},
  archiveId: null,
  rightPanelActiveTab: 'closed',
  toggleParticipantList: () => {},
  toggleBackgroundEffects: () => {},
  toggleChat: () => {},
  closeRightPanel: () => {},
  toggleReportIssue: () => {},
  pinSubscriber: () => {},
  isMaxPinned: false,
  ownCaptions: null,
  sendEmoji: () => {},
  emojiQueue: [],
  publish: async () => Promise.resolve(),
  unpublish: () => {},
  lastStreamUpdate: null,
  subscriptionError: null,
});

// GOOD: forces consumers to be inside the provider, or fail loudly
export const SessionContext = createContext<SessionContextType | null>(null);
```

Required follow-up (to avoid runtime null errors and to improve diagnostics):

- Update the session context consumer(s) to enforce provider presence. For example, a `useSessionContext` hook that throws a clear error when the provider is missing.
- Adjust tests accordingly so they fail when the provider is not mounted, instead of silently using default no-op behavior.

This change will directly address why removing the wrapper has no effect, and it will make missing provider usage detectable during tests and during the migration.

... also just like a comment, this test is testing only JSX not functionality, that's not what we want but I understand the .spec was already there

docs/TESTING.md
johnny-quesada-developer, line 1: Hi @Hossein-Movahed , after attending the other feedback please update this readme Ill take a look then

frontend/src/test/fixtures/userFixtures.ts
```diff
@@ -0,0 +1,92 @@
+import { UserType } from '@Context/user';
+
+/**
+ * Default user fixture for testing.
+ * Provides a complete user object with sensible defaults.
+ */
+export const defaultUserFixture: UserType = {
```
johnny-quesada-developer: Please follow the code guidelines on [single-responsibility-principle--file-organization](https://github.com/Vonage/vonage-video-react-app/blob/develop/.github/copilot-instructions.md#single-responsibility-principle--file-organization).

In general we don't want to have barrel files... And specifically for unit tests, we should intentionally minimize shared abstractions. For unit tests, prefer explicit, per-test setup unless there is very strong reason like with the provider.

johnny-quesada-developer (changes requested):

frontend/src/test/mocks/mediaDevices.ts
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * Mock implementation of navigator.mediaDevices for testing.
+ * Provides realistic browser media API behavior without actual device access.
+ */
+
+/**
+ * Creates a mock MediaDevices object.
+ */
+export function createMockMediaDevices() {
```
johnny-quesada-developer: Hi @Hossein-Movahed , 

I understand the intention of this kind of helpers, but right now there is no need of adding it...

Talking specifically about the media devices, we should not consume them directly, and a helper like we have right now is not enough because is not reactive. We'll be migrating the media devices to use context (SOON), and then like with whatever other of the context we have, you'll be able to modify the context state per test to match each specific use case. 

Then about unit test in general: 
- We want to keep abstraction at minimum.
- Each test should be isolated, and explicit.

```tsx 
// Bad
beforeEach(() => {
  /// if removed or modified could potentially break other tests, NOT ISOLATED 
  setupMockMediaDevices();
})
```

Instead we want each test to be simple and deterministic: 

```tsx
it('should test something....', async () => {
  const { MediaDevicesWrapper } = makeMediaDevicesWrapper({
    mediaDevices: {
      enumerateDevices: async () => [
        { deviceId: 'audio1', kind: 'audioinput', label: 'Microphone 1' },
        { deviceId: 'video1', kind: 'videoinput', label: 'Camera 1' },
      ],
    },
  });

  const { container } = render(<MyComponent />, { wrapper: MediaDevicesWrapper })
 
  expect(container.querySelectorAll('option').length).toBe(2)
})

it('should test something....', async () => {
  jest.mocked(navigator.mediaDevices).mockReturnValue({
    enumerateDevices: async () => [
      { deviceId: 'audio1', kind: 'audioinput', label: 'Microphone 1' },
      { deviceId: 'video1', kind: 'videoinput', label: 'Camera 1' },
    ],
  });

  const { container } = render(<MyComponent />)

  expect(container.querySelectorAll('option').length).toBe(2)
})
```

We could have a Navigator Mock, but it should not magically resolve the test use-cases:

```tsx
export class NavigatorMock implements Navigator {
  mediaDevices: {
    getUserMedia: jest.fn(() => {
      throw new Error('navigator.mediaDevices.getUserMedia not implemented');
    }),
    enumerateDevices: jest.fn(() => {
      throw new Error('navigator.mediaDevices.enumerateDevices not implemented');
    }),
  },
  ....

jest.beforeAll(() => {
  global.navigator = new NavigatorMock();
});  
```

frontend/src/test/mocks/vonageVideo.ts
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * Mock implementation of the Vonage Video SDK.
+ * Provides realistic mocks for testing without external dependencies.
+ */
+
+/**
+ * Creates a mock Publisher instance with common methods.
+ */
+export function createMockPublisher(overrides: Partial<Publisher> = {}): Publisher {
```
johnny-quesada-developer: We probably dont want this mock neither, I'm almost sure there are functionalities and helpers of **vonageVideo** who work perfectly even during test, so why mocking them?

Mocking should be minimum, for example just enough to reproduce an specific use case, or to overpass and environment limitation. 

 So, as with the previous [example](https://github.com/Vonage/vonage-video-react-app/pull/300/files#r2619058737), we can have something like:

 ```tsx
export class SessionMock extends Session {
  publish = jest.fn(() => {
    return throw new Error('You need to mock publish() in your test');
  });
  ....

jest.mock('@vonage/video-client', () => {
  const originalModule = jest.requireActual('@vonage/video-client');

  return {
    __esModule: true,
    ...originalModule,
    Session: jest.fn().mockImplementation(() => {
      return new SessionMock();
    }),
  };
});
```

File: frontend/src/test/providers/makeAudioOutputProviderWrapper.ts [hunk 7]
```diff
@@ -0,0 +1,25 @@
+import { AudioOutputProvider, AudioOutputContext } from '@Context/AudioOutputProvider';
+import makeGenericProviderWrapper, { GenericWrapperOptions } from './makeGenericProviderWrapper';
+
+export type AudioOutputProviderWrapperOptions = GenericWrapperOptions<
```
johnny-quesada-developer: Hi @Hossein-Movahed , as we commented previously, even for providers who looks simple and right now they do not have dependencies lets use this structure instead:
```tsx
export type AudioOutputProviderWrapperOptions = {
  audioOutputOptions?: GenericWrapperOptions<
   typeof AudioOutputProvider,
   typeof AudioOutputContext
  >;
};
```

frontend/src/test/providers/makeBackgroundPublisherProviderWrapper.ts
```diff
@@ -0,0 +1,27 @@
+import {
+  BackgroundPublisherProvider,
+  BackgroundPublisherContext,
+} from '@Context/BackgroundPublisherProvider';
+import makeGenericProviderWrapper, { GenericWrapperOptions } from './makeGenericProviderWrapper';
+
+export type BackgroundPublisherProviderWrapperOptions = GenericWrapperOptions<
```
johnny-quesada-developer: [commend](https://github.com/Vonage/vonage-video-react-app/pull/300/files#r2619127382)

frontend/src/test/providers/makePreviewPublisherProviderWrapper.ts
```diff
@@ -0,0 +1,28 @@
+import {
+  PreviewPublisherProvider,
+  PreviewPublisherContext,
+} from '@Context/PreviewPublisherProvider';
+import makeGenericProviderWrapper, { GenericWrapperOptions } from './makeGenericProviderWrapper';
+
+export type PreviewPublisherProviderWrapperOptions = GenericWrapperOptions<
```
johnny-quesada-developer: [commend](https://github.com/Vonage/vonage-video-react-app/pull/300/files#r2619127382)

frontend/src/test/providers/makePublisherProviderWrapper.ts
```diff
@@ -0,0 +1,25 @@
+import { PublisherProvider, PublisherContext } from '@Context/PublisherProvider';
+import makeGenericProviderWrapper, { GenericWrapperOptions } from './makeGenericProviderWrapper';
+
+export type PublisherProviderWrapperOptions = GenericWrapperOptions<
```
johnny-quesada-developer: [commend](https://github.com/Vonage/vonage-video-react-app/pull/300/files#r2619127382)

frontend/src/test/providers/makePublisherProviderWrapper.ts
```diff
@@ -0,0 +1,25 @@
...
+>;
+
+/**
+ * Creates wrapper for the PublisherProvider context.
+ * Allows accessing the context value for testing.
+ * @param options - The wrapper options.
+ * @returns The PublisherProvider wrapper and context getter.
+ */
+function makePublisherProviderWrapper(options?: PublisherProviderWrapperOptions) {
+  const [PublisherProviderWrapper, publisherContext] = makeGenericProviderWrapper(
```
johnny-quesada-developer, line 43: Hi @Hossein-Movahed ,

Please review how we implement wrappers for contexts that depend on each other. The **PublisherContext** depends on the **SessionContext**, and the **SessionContext** depends on the **UserContext** and **AppConfigContext**. In your current implementation, none of the related contexts are taken into account or accessible for reading or writing.

johnny-quesada-developer (changes requested):

frontend/src/test/mocks/index.ts
johnny-quesada-developer, line 1: @Hossein-Movahed , delete empty file

frontend/src/pages/MeetingRoom/MeetingRoom.spec.tsx
```diff
@@ -25,7 +25,12 @@ import useToolbarButtons, {
   UseToolbarButtonsProps,
 } from '@hooks/useToolbarButtons';
 import usePublisherOptions from '@Context/PublisherProvider/usePublisherOptions';
-import { makeAppConfigProviderWrapper } from '@test/providers';
+import {
```
johnny-quesada-developer: You don't need to import all the wrappers, just the lowest level which its already using all the others, in this case usePublisher.
```tsx
import { makePublisherProviderWrapper, PublisherProviderWrapperOptions } from '@test/providers';
```

Im attaching the example test:
```tsx
// mock new VonageVideoClient() // return mock and track connection

it.only('should call publish after connected', () => {
  const { sessionContext, publisherContext } = render(<MeetingRoom />);

  expect(sessionContext.current.joinRoom).toHaveBeenCalledWith('test-room-name');

  waitFor(() => {
    expect(publisherContext.current.publish).toHaveBeenCalledTimes(1);
  });
});

// mock 'opentok-layout-js' default.getLayout
it.only('should display publisher', () => {
  const { rerender } = render(<MeetingRoom />, {
    publisherContext: {
      initialValue: {
        publisher: mockPublisher,
        isPublishing: true,
      },
    },
    sessionContext: {
      // todo, allow initial value to be set here
      __onCreated: (context) => {
        context.connected = true;
      },
    },
  });

  rerender(<MeetingRoom />);

  waitFor(() => {
    expect(screen.getByTestId('publisher-container')).toBeInTheDocument();
  });
});
```

johnny-quesada-developer (changes requested):

frontend/src/pages/MeetingRoom/MeetingRoom.spec.tsx
```diff
@@ -254,125 +293,241 @@ describe('MeetingRoom', () => {
...
+      act(() => {
+        publisherContext.current.isVideoEnabled = false;
+      });
+
+      act(() => {
+        rerender(<MeetingRoom />);
+      });
+
+      waitFor(() => {
+        const connectionAlert = screen.queryByText(
```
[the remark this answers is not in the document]
Hossein-Movahed: We are skipping this test

frontend/src/pages/MeetingRoom/MeetingRoom.spec.tsx
```diff
@@ -254,125 +293,241 @@ describe('MeetingRoom', () => {
...
+          },
+        },
+      });
+
+      act(() => {
+        rerender(<MeetingRoom />);
+      });
+
+      waitFor(() => {
+        const connectionAlert = screen.queryByText(
```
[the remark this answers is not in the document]
Hossein-Movahed: We are skipping this test

frontend/src/test/helpers/renderWithProviders.ts
```diff
@@ -0,0 +1,98 @@
...
+ * Renders a component with composed AppConfig and AudioOutput providers
+ */
+export function renderWithAppConfigAndAudioOutput(
+  ui: ReactElement,
+  options?: {
+    appConfigOptions?: AppConfigProviderWrapperOptions;
+    audioOutputOptions?: AudioOutputProviderWrapperOptions['audioOutputOptions'];
+  }
+) {
+  const { AppConfigWrapper } = makeAppConfigProviderWrapper(options?.appConfigOptions);
```
johnny-quesada-developer:
```suggestion
  const { AppConfigWrapper, AppConfigContext } = makeAppConfigProviderWrapper(options?.appConfigOptions);
```

johnny-quesada-developer (changes requested):

OscarFava (dismissed):

OscarFava (dismissed):

frontend/src/Context/AudioOutputProvider/useAudioOutput/useAudioOutput.spec.tsx
```diff
@@ -1,75 +1,59 @@
...
 
 vi.mock('@vonage/client-sdk-video');
 
 describe('useAudioOutput', () => {
-  const nativeMediaDevices = global.navigator.mediaDevices;
   let mockGetActiveAudioOutputDevice: MockInstance<[], Promise<AudioOutputDevice>>;
   let mockSetAudioOutputDevice: MockInstance<[deviceId: string], Promise<void>>;
 
   beforeEach(() => {
     vi.resetAllMocks();
```
johnny-quesada-developer: you dont need to call this... we already cleaning all the mocks on the project setup

frontend/src/Context/BackgroundPublisherProvider/useBackgroundPublisher/useBackgroundPublisher.spec.tsx
```diff
@@ -158,9 +130,7 @@ describe('useBackgroundPublisher', () => {
 
     it('clears video filter for unknown option', async () => {
       await act(async () => {
```
johnny-quesada-developer, line 146: hi @Hossein-Movahed,

Where are the expects here? What are we testing?

frontend/src/Context/AudioOutputProvider/useAudioOutput/useAudioOutput.spec.tsx
```diff
@@ -1,75 +1,59 @@
...
-      value: {
-        enumerateDevices: vi.fn(
-          () =>
-            new Promise<MediaDeviceInfo[]>((res) => {
-              res(nativeDevices as MediaDeviceInfo[]);
-            })
-        ),
-        addEventListener: vi.fn(() => []),
-        removeEventListener: vi.fn(() => []),
+    setupNavigatorMocks({
```
johnny-quesada-developer: Hi @Hossein-Movahed,

These kinds of mocks that magically make tests pass without reflecting the actual implementation are exactly what we are trying to eliminate.

The current approach adds dummy functions to the mediaDevices API without requiring an understanding of how the API behaves. As a result, the tests omit the behaviors we actually rely on and need to verify.

useAudioOutput must register a devicechange event listener. At the moment, this is not tested at all because the mock makes it invisible. Similarly, when the component unmounts, the listener must be removed. Both behaviors are part of the contract and should be covered by tests to ensure compatibility over time.

We already have a simple, reusable mock that makes these side effects explicit:

```tsx
const mediaDevices: typeof navigator.mediaDevices = {
  ondevicechange: null,
  enumerateDevices() {
    throw new Error('Function not implemented.');
  },
  getDisplayMedia() {
    throw new Error('Function not implemented.');
  },
  getSupportedConstraints() {
    throw new Error('Function not implemented.');
  },
  getUserMedia() {
    throw new Error('Function not implemented.');
  },
  addEventListener() {
    throw new Error('Function not implemented.');
  },
  removeEventListener() {
    throw new Error('Function not implemented.');
  },
  dispatchEvent() {
    throw new Error('Function not implemented.');
  },
};
```

We can register this mock globally, but tests still need to interact with it explicitly. The goal is to observe what the component does and assert on the side effects. Adding empty helpers that we neither control nor spy on puts us in the same position as before: tests that pass without validating behavior.

For example:
```tsx
it('should initialize the store correctly', async () => {
  vi.spyOn(mediaDevicesMock, 'addEventListener').mockImplementation(() => {});
  vi.spyOn(mediaDevicesMock, 'enumerateDevices').mockResolvedValue(mediaDevices);

  const { result } = renderHook(() => devices$());

  const [state, api] = result.current;

  expect(state.audioOutputDevices).toBeDefined();
  expect(state.devices).toBeDefined();
  expect(api.getConnectedDeviceId).toBeDefined();
  expect(api.updateMediaDevices).toBeDefined();

  expect(mediaDevicesMock.addEventListener).toHaveBeenCalledWith(
    'devicechange',
    expect.any(Function)
  );

  expect(mediaDevicesMock.enumerateDevices).toHaveBeenCalled();
});

```

This was intended to be a small, mechanical change, but it has grown because the same issue keeps reappearing. The core problem we are trying to fix is over-mocking that hides behavior. Helpers that abstract mocks and make tests pass without asserting side effects are part of that problem, not a solution.

Please remove the helper-based mock and keep the mediaDevices mock explicit in the test, with assertions for the observable side effects (devicechange registration and cleanup). 

cc: @OscarFava

frontend/src/Context/AudioOutputProvider/useAudioOutput/useAudioOutput.spec.tsx
```diff
@@ -1,75 +1,59 @@
-import { beforeEach, describe, it, expect, vi, afterAll, MockInstance } from 'vitest';
-import { act, renderHook, waitFor } from '@testing-library/react';
+import { beforeEach, describe, it, expect, vi, MockInstance } from 'vitest';
+import { act, waitFor } from '@testing-library/react';
 import { AudioOutputDevice } from '@vonage/client-sdk-video';
 import * as OT from '@vonage/client-sdk-video';
+import { setupNavigatorMocks } from '@test/setup/setupNavigatorMocks';
+import { renderHookWithAudioOutput as render } from '@test/helpers/renderWithProviders';
```
johnny-quesada-developer: Most of the setup in renderWithProviders is only used by a single test file. Please move that setup into the test itself instead of extending a shared helper.

Also, our [code guidelines](https://github.com/Vonage/vonage-video-react-app/blob/develop/.github/copilot-instructions.md#folder-and-file-structure) explicitly state that we do not want barrel files. Let’s avoid introducing them here and keep the structure aligned with the guidelines.

frontend/src/Context/AudioOutputProvider/useAudioOutput/useAudioOutput.spec.tsx
```diff
@@ -1,75 +1,59 @@
...
   });
 
   it('should call getActiveAudioOutputDevice when initialized', async () => {
-    const { result } = renderHook(() => useAudioOutput());
+    const { result } = render(useAudioOutput);
 
     await waitFor(() => expect(result.current.currentAudioOutputDevice).toBe('some-device-id'));
 
-    expect(mockGetActiveAudioOutputDevice).toHaveBeenCalledOnce();
+    expect(mockGetActiveAudioOutputDevice).toHaveBeenCalled();
```
johnny-quesada-developer: why are we downgrating the specificcity of the test?

frontend/src/Context/BackgroundPublisherProvider/useBackgroundPublisher/useBackgroundPublisher.spec.tsx
```diff
@@ -43,32 +23,24 @@ describe('useBackgroundPublisher', () => {
 
   const mockedInitPublisher = vi.fn();
   const mockedHasMediaProcessorSupport = vi.fn();
-  const mockSetAccessStatus = vi.fn();
 
   beforeEach(() => {
     vi.spyOn(console, 'error');
-    vi.mocked(useUserContext).mockImplementation(() => mockUserContextWithDefaultSettings);
+
+    setupNavigatorMocks();
```
johnny-quesada-developer: setupNavigatorMocks introduces the same over-mocking we are trying to clean up. Please remove it and explicitly mock and assert only the window.navigator APIs this code actually uses.

This was intended to be a mechanical task, but it’s taking longer because the same pattern keeps being reintroduced. Let’s keep this change focused and aligned with the goal of the cleanup.

cc: @OscarFava

frontend/src/Context/BackgroundPublisherProvider/useBackgroundPublisher/useBackgroundPublisher.spec.tsx
```diff
@@ -193,61 +163,77 @@ describe('useBackgroundPublisher', () => {
...
+      setupNavigatorMocks({
+        permissions: {
           query: mockQuery,
         },
       });
     });
 
     afterAll(() => {
-      Object.defineProperty(global.navigator, 'permissions', {
+      Object.defineProperty(globalThis.navigator, 'permissions', {
```
johnny-quesada-developer: By using regular vitest mocks this is probably not necessary...

Every function overrided with vi.spy or vi.mockImplementation gets restored after each test due our setup

frontend/src/Context/PublisherProvider/usePublisher/usePublisher.spec.tsx
```diff
@@ -120,12 +120,15 @@ describe('usePublisher', () => {
 
   describe('changeBackground', () => {
     let result: ReturnType<typeof renderHook>['result'];
-    beforeEach(() => {
+    beforeEach(async () => {
       vi.mocked(initPublisher).mockImplementation(() => mockPublisher);
       result = renderHook(() => usePublisher()).result;
       act(() => {
         (result.current as ReturnType<typeof usePublisher>).initializeLocalPublisher({});
```
johnny-quesada-developer: Hi @Hossein-Movahed, please remove this global result. Tests should be isolated, and global variables in test files are an antipattern.

frontend/src/Context/SessionProvider/session.spec.tsx
```diff
@@ -125,7 +125,7 @@ describe('SessionProvider', () => {
     );
   };
 
-  beforeEach(() => {
+  beforeEach(async () => {
```
johnny-quesada-developer: This **beforeEach** is acting as a test harness (state, mocks, render, assertions); please move instance creation and assertions into the individual tests to keep them isolated and readable.

frontend/src/components/HiddenParticipantsTile/HiddenParticipantsTile.spec.tsx
```diff
@@ -122,16 +123,12 @@ describe('HiddenParticipantsTile', () => {
 });
 
 function render(ui: ReactElement, options?: SessionProviderWrapperOptions) {
-  const { SessionProviderWrapper, sessionContext, appConfigContext } = makeSessionProviderWrapper({
-    __onCreated: (sessionContext) => {
-      vi.spyOn(sessionContext, 'toggleParticipantList');
+  return renderWithSession(ui, {
```
johnny-quesada-developer: I don’t see the value of this wrapper. It doesn’t remove any real complexity and only adds an extra indirection to the test setup.

frontend/src/components/MeetingRoom/DeviceControlButton/DeviceControlButton.spec.tsx
```diff
@@ -1,13 +1,13 @@
...
 import { Publisher } from '@vonage/client-sdk-video';
 import { EventEmitter } from 'stream';
-import { ReactElement } from 'react';
 import { PublisherContextType } from '@Context/PublisherProvider';
 import useSpeakingDetector from '@hooks/useSpeakingDetector';
 import usePublisherContext from '@hooks/usePublisherContext';
 import { defaultAudioDevice } from '@utils/mockData/device';
-import { AppConfigProviderWrapperOptions, makeAppConfigProviderWrapper } from '@test/providers';
+import { setupNavigatorMocks } from '@test/setup/setupNavigatorMocks';
+import { renderWithAppConfig as render } from '@test/helpers/renderWithProviders';
```
johnny-quesada-developer: This **renderWithAppConfig** looks like unnecessary indirection. It’s better to keep the render logic explicit and simple unless it’s strictly necessary.

frontend/src/test/helpers/renderWithProviders.ts
johnny-quesada-developer, line 1: Hi @Hossein-Movahed,

Let’s focus on the actual goal of this work. We want to remove over-mocking and make tests more explicit and reliable. Introducing a whole new layer of test helpers and barrel-style abstractions hides behavior and goes against what we discussed.

Going forward, the intention is to keep things simple and explicit ahead of the upcoming refactors. The goal is to ensure the tests actually verify behavior and fail when that behavior breaks.

johnny-quesada-developer: The concern here is not duplication or code style. It’s that test abstractions tend to grow until the test suite becomes its own project. At that point, teams end up spending more time fixing tests and maintaining complex abstractions than validating real behavior in the code under test.

We intentionally want to keep tests simple and explicit to avoid that outcome.

frontend/src/test/setup.ts
```diff
@@ -7,6 +7,31 @@ import '../i18n';
 // Mock scrollIntoView for jsdom environment
 Element.prototype.scrollIntoView = vi.fn();
 
+Object.defineProperty(HTMLMediaElement.prototype, 'play', {
```
johnny-quesada-developer, line 10: These global mocks make the environment “work” but also make failures invisible. Can we make them strict by default (throw with a clear message) so tests are forced to explicitly mock/verify behavior instead of passing via no-op polyfills?

```tsx
Object.defineProperty(HTMLMediaElement.prototype, 'play', {
  configurable: true,
  value: vi.fn(() => {
    throw new Error(
      'HTMLMediaElement.play() was called during a test. ' +
      'This method is not implemented in jsdom. ' +
      'Mock or spy on it explicitly and assert the expected behavior.'
    );
  }),
});
```

frontend/src/test/setup/setupNavigatorMocks.ts
johnny-quesada-developer, line 1: This helper hides which navigator APIs are actually used and provides green-by-default mocks.

As a result, tests can pass without verifying side effects (for example, devicechange listener registration and removal). This introduces the same problem we had with mocked contexts.

Please inline a strict mock per test (or use a strict global that throws) and explicitly assert the expected calls and behavior.

frontend/src/utils/publisher/handlePublisherAccessDenied.ts
johnny-quesada-developer, line 1: Nice catch on the duplication and extracting this logic into a single place. This makes the intent clearer and avoids repeating the same permission-handling code.

We’ll probably want to move this to better align with our [domain structure](https://github.com/Vonage/vonage-video-react-app/blob/develop/.github/copilot-instructions.md#folder-and-file-structure) later on, but this is a good improvement.

sonar-project.properties
```diff
@@ -30,6 +30,7 @@ customWordList.mjs,\
 **/vite.config.*,\
 **/webpack.config.*,\
 libs/ui/src/theme/helpers/designTokens/**
+frontend/src/test/mocks/vonageVideo.ts
```
johnny-quesada-developer: you dont need to completelly exclude the file from the analysis, if the problem is test coverage you can use:
```
sonar.coverage.exclusions=
```

johnny-quesada-developer (changes requested):
hi @Hossein-Movahed ,

Thanks for the hard work on this. The original goal of this PR was a relatively simple cleanup focused on reducing over-mocking and making tests more explicit and reliable.

A few changes go in a different direction by introducing additional abstractions and helpers. I’ve left inline comments where adjustments are needed to bring things back in line with that original intent.

Let me know once you’ve addressed the feedback so we can merge these changes and move forward.

frontend/src/Context/AudioOutputProvider/useAudioOutput/useAudioOutput.spec.tsx
Hossein-Movahed: You're right that some of these helpers are currently only used by one or two test files. This PR lays the foundation for refactoring test setup across the codebase, I've updated a few test files as examples (the ones that were failing plus a couple of samples).

My plan is:
Complete the broader test refactor to see which helpers are actually reused across multiple tests
Remove/inline any helpers that end up only being used in a single place
Keep only the truly shared helpers that provide value

I can either:
Continue with this approach and clean up unused helpers once we see the usage patterns, or
Remove the rarely-used helpers now and have tests set up their contexts inline

What do you think?

frontend/src/components/HiddenParticipantsTile/HiddenParticipantsTile.spec.tsx
Hossein-Movahed: I see your point on the indirection, but I think the wrapper helps here:

1. Keeps tests cleaner, just `render()` instead of repeating the `__onCreated` spy setup in all 3 tests
2. Consistent with other test files in the codebase
3. If we need to change the spy setup later, one place to update

It's only ~8 lines and removes duplication sonar error. Worth keeping imo, but happy to hear your thoughts.

frontend/src/components/MeetingRoom/DeviceControlButton/DeviceControlButton.spec.tsx
Hossein-Movahed: Good point, the `as render` alias was hiding what it actually does. I removed the alias so now it's explicit that we're `using renderWithAppConfig`.

The helper itself is still useful though, the component depends on `AppConfig` context, and `renderWithAppConfig` just wraps it with that provider. Without it, we'd need to manually wrap every render call with the provider, which would be more boilerplate

frontend/src/test/helpers/renderWithProviders.ts
Hossein-Movahed: I created these helpers to reduce duplication for `SonarQube`. I see your concern about test abstractions though.

If you prefer more explicit tests, I can remove the helpers and we can adjust `SonarQube`'s duplication threshold for test files.

One thing to consider:  I haven't removed context mocking everywhere yet. When tests need multiple providers (`Session` + `AppConfig` + `AudioOutput`, etc.), composing them manually in each test gets complex quickly. The helpers simplify that, but I'm open to your preference on handling it. 
Let me know

frontend/src/test/setup/setupNavigatorMocks.ts
Hossein-Movahed: I replaced `setupNavigatorMocks` with explicit per-file mocks like you asked, but it pushed duplication to 6.4% (limit is 3%).

I extracted two tiny helpers to reduce it:
`createMediaDevicesMock`,`setupMediaDevicesMock`

That gets us down to 3.6% (still slightly over). Mocks are still explicit in each beforeEach.

Not a huge fan of the `setupMediaDevicesMock` either 😅

Options:

- increase the %
- Exclude tests from duplication
- Extract more (less explicit)

What do you think?

frontend/src/Context/AudioOutputProvider/useAudioOutput/useAudioOutput.spec.tsx
johnny-quesada-developer: Hi @Hossein-Movahed ,

Thanks for the explanation. I don’t agree with introducing a helper layer first and cleaning it up later. We should only add abstractions when we are confident they are needed; otherwise, we risk investing time and effort in something that provides little or no value.

The original scope of this PR was to reduce over-mocking and make tests more explicit and reliable. Adding **@test/helpers/*** and **green-by-default** setup (such as **setupNavigatorMocks** and **renderWithProviders**) moves in the opposite direction. It hides which **APIs** and providers are actually required and allows tests to pass without asserting the side effects we care about. This is effectively a different form of **over-mocking** and introduces unnecessary complexity.

These helpers are currently used by only one or two tests. That’s not reuse yet, it’s indirection. It adds an additional abstraction layer, implicit dependencies, long-term maintenance cost, and it works against the goal of keeping tests simpler and more reliable.

**What has consistently proven to work better, and aligns with common unit testing best practices, is:**

- keeping mocks local to the test file that uses them
- explicit provider wiring at the test level

This keeps dependencies visible, avoids hidden behavior, and prevents unintended coupling between tests.

Helpers are not banned, but we should avoid building a test framework. Abstractions should be introduced only when they are strictly necessary, for example provider wrappers that increase context observability and accessibility rather than hiding it.

To summarize: we should keep tests explicit, avoid **green-by-default** mocks, and avoid helpers that reduce observability and accessibility of the behavior under test.

### Also, **Small process note**: 
review comments should generally be left open and resolved by the reviewer once the feedback is addressed. This helps keep to track of the review process.

cc: @OscarFava

Hossein-Movahed: Totally agree with you! I only created these helpers to fix the SonarQube duplication error (6.4% → 3.6%, but still over the 3%).

I'd much rather keep everything explicit and local per test file like you described. If you're good with excluding tests from duplication checks (`sonar.cpd.exclusions`), I'll remove the helpers and inline everything.

Noted about the resolving comment, didnt know

johnny-quesada-developer (approved):

johnny-quesada-developer (approved):
LGTM

OscarFava (approved):

PR #345, opened 2026-01-23
VIDSOL-465: Add Recording & Transcription Indication 

behei-vonage:
#### What is this PR doing?

This PR adds a Recording & Transcription Indication when one person starts a recording and gives an option to either remain in the session or leave. 

Checkout this branch.
Join the meeting room with two publishers. 
Start the recording from one, notice the popup in the other tab once started. 
Do not consent to the recording on that tab, notice that you are redirected to the goodbye page without an option provided to download archives. 

Resolves [VIDSOL-465](https://jira.vonage.com/browse/VIDSOL-465)

#### Checklist
[ ] Resolves a `Known Issue`.


OscarFava (dismissed):
Awesome feature!

LGTM!

OscarFava (dismissed):

Hossein-Movahed (dismissed):
LGTM 🔥

sharad-srivastava2:
Quick update after testing against the new acceptance criteria from https://jira.vonage.com/browse/VIDSOL-465

Overall, the implementation aligns well with the updated flow where participants are not asked for explicit consent, but are notified and can choose to remain or leave.

✅ Recording start notification, existing participant handling, late joiner notification, and rejoin-after-decline scenarios are all working as expected.

⚠️ One gap noticed: in the Recording Restart flow (Start → Stop → Start), participants are notified again with the consent popup, whereas the acceptance criteria expect the recording to resume without re-requesting consent, and only show a notification like “Recording has resumed”.

frontend/src/Context/SessionProvider/session.tsx
Hossein-Movahed, line 134: `recordingAlreadyNotified` and `archiveIdStartedBySelf` should be added to `SessionContextInitialValue`.

frontend/src/Context/SessionProvider/session.tsx
```diff
@@ -169,6 +179,17 @@ const SessionProvider = ({ children, initialValue = {} }: SessionProviderProps):
   );
 
   const [archiveId, setArchiveId] = useState<string | null>(initialValue?.archiveId ?? null);
+  const [archiveIdStartedBySelf, setArchiveIdStartedBySelf] = useState<string | null>(null);
```
Hossein-Movahed: They should then be initialized using useState, like:
`useState<string | null>(initialValue.archiveIdStartedBySelf)` and
`useState<boolean>(initialValue.recordingAlreadyNotified)`.

Hossein-Movahed (commented):
Amazing work on this feature really cool! 🔥 

I just have a small comment regarding the new states. Also, while testing the flow locally, I noticed that if the second participant rejects the recording consent, they are redirected to the goodbye page as expected. However, the final archive still contains a few seconds of their video before the redirect happens.

Is this expected behavior due to the archive starting immediately when recording is triggered? Just wondering if this is something we should be aware of from a privacy or UX perspective.

OscarFava (dismissed):

behei-vonage:
> Amazing work on this feature really cool! 🔥
> 
> I just have a small comment regarding the new states. Also, while testing the flow locally, I noticed that if the second participant rejects the recording consent, they are redirected to the goodbye page as expected. However, the final archive still contains a few seconds of their video before the redirect happens.
> 
> Is this expected behavior due to the archive starting immediately when recording is triggered? Just wondering if this is something we should be aware of from a privacy or UX perspective.

great catch, @Hossein-Movahed! 
I've actually been thinking about this one since you mentioned it. I think the ideal scenario would be to track the state of each connection and only start the archiving after a response has been received. But this also will introduce an unnecessary step and crucial parts of recording might be missed.

Instead, I added a 3 second delay before actually starting the archiving in the `ArchivingButton` component. This is the most straight-forward way in my opinion as user should give their consent / reject and leave within the first three seconds.

VZaphod:
@Hossein-Movahed @behei-vonage I think that it's ok if there are some seconds of recording, we wont gather consent from everybody before starting the recording, just letting know to others that the meeting is going to be recorded with the possibility of leaving the meeting.

The specified acceptance criteria states:
**Given** a new recording started **When** the rest of the users are notified **Then** the notified user can remain or leave.

Hossein-Movahed (dismissed):
LGTM 🔥

OscarFava (dismissed):
nice work 💯

sharad-srivastava2:
Good discussion, and it makes sense @behei-vonage @Hossein-Movahed @OscarFava @VZaphod 👍
One thing I’d like to clarify, what happens in the case where a participant does not take any action on the notification (neither accepts nor declines)?
Do we have any enforcement in place for that scenario, or can the user continue staying in the call without explicitly choosing to remain or leave?
Just thinking from a compliance perspective, since this could be a grey area if the user is effectively recorded without taking any action.

Also, Update: I re-tested the Recording Restart scenario, and this looks good now 👍 The behaviour aligns with the acceptance criteria and recording resumes without re-requesting consent, and participants are only notified appropriately.

sharad-srivastava2 (dismissed):

behei-vonage:
Note: SonarCloud Code Analysis failed for reasons not related to this PR:

image

Hossein-Movahed (approved):

OscarFava (approved):

PR #357, opened 2026-02-04
VIDSOL-488: firefox video preview

Hossein-Movahed:
#### What is this PR doing?
**Context**
Firefox may return an empty string ("") for `deviceId` before camera/mic permission is granted.
Passing this value to `OpenTok` publisher constraints causes `getUserMedia` to fail with `OverconstrainedError`, crashing the app and preventing the local preview from loading.

**Fix**
Only pass `audioSource` / `videoSource` when a valid `deviceId` is available.
When omitted, `OpenTok` correctly falls back to the default devices, allowing Firefox to initialize safely.

Run the project in Firefox or Safari, go to a waiting room, and check that the video preview loads without crashing.

Resolves [VIDSOL-488](https://jira.vonage.com/browse/VIDSOL-488)

#### Checklist
[ ] Resolves a `Known Issue`.


OscarFava (dismissed):
It works on safari and firefox for me! LGTM!

frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.tsx
```diff
@@ -229,12 +229,13 @@ const usePreviewPublisher = (
...
     }
 
+    // Omit empty deviceId so SDK uses default device (Firefox returns "" before permission).
     const publisherOptions: PublisherProperties = {
       insertDefaultUI: false,
       videoFilter,
       resolution: defaultResolution,
-      audioSource,
-      videoSource,
+      ...(audioSource && { audioSource }),
```
johnny-quesada-developer: Hi @Hossein-Movahed ,

**PublisherProperties** accepts undefined value so lets just avoid sending empty strings... 

```tsx
      audioSource: audioSource || undefined,
      videoSource: videoSource || undefined,,
```

and we have the same problem on:
**./frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/useBackgroundPublisher.tsx**

johnny-quesada-developer (changes requested):

OscarFava (dismissed):

OscarFava (dismissed):

OscarFava (dismissed):

frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.tsx
```diff
@@ -239,12 +239,13 @@ const usePreviewPublisher = (
...
     }
 
+    // Omit empty deviceId so SDK uses default device (Firefox returns "" before permission).
     const publisherOptions: PublisherProperties = {
       insertDefaultUI: false,
       videoFilter,
       resolution: defaultResolution,
-      audioSource: audioSourceId,
-      videoSource: videoSourceId,
+      audioSource: audioSourceId || undefined,
```
johnny-quesada-developer: Hi @Hossein-Movahed , 

The current version of develop already takes care of that... as we saw in the call this is not the reason why the publisher is failing.  We need to check on the code of the package, what is making the publisher fail and tackle the actual problem. 

cc: @OscarFava

johnny-quesada-developer (changes requested):

frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.tsx
Hossein-Movahed: I checked the `develop` branch. Develop passes `audioSource` and `videoSource` directly to the publisher with no special handling:
`audioSource: audioSourceId,`
` videoSource: videoSourceId,`

In Firefox, before permission is granted, `enumerateDevices()` can return devices with `deviceId: ""`. Our `reviseMediaSelection` can then set `audioinput`/`videoinput` to `""`, so we end up passing `audioSource: ""` and `videoSource: ""` to the SDK.

MDN and the Media Capture spec say that passing an empty or invalid deviceId to `getUserMedia` can cause `OverconstrainedError` because no device matches that constraint. So this does look like a valid cause of the failure.

The fix is to omit `audioSource` and `videoSource` when they’re `falsy` so the SDK doesn’t pass invalid device constraints to `getUserMedia`. If you’re seeing a different root cause in the SDK, can you share what you found?

johnny-quesada-developer (changes requested):
There are two core issues:

- Devices appear empty in Firefox.
- There is a lack of synchronization once the Vonage SDK successfully retrieves the corresponding permissions.

This PR should not proceed as it stands. Although we added some filtering for device IDs, it was implemented in the wrong place. Additionally, the PR does not include the necessary synchronization fixes.

johnny-quesada-developer:
will fix with [this PR](https://github.com/Vonage/vonage-video-react-app/pull/373)

PR #366, opened 2026-02-12
VIDSOL-509: add backend Gollum wireup for client events

Hossein-Movahed:
#### What is this PR doing? 
Implements the backend logging provider. Client events (CallStarted, CallEnded, errors, etc.) are sent from the frontend to the backend, which forwards them to Gollum/HLG for Kibana analytics.

- add `GOLLUM_BASE_URL` to your env file 
### Backend
- **Route**: `POST /internal/client-logs` with 50kb JSON body limit
- **gollumClientService**: Forwards validated events to Gollum when `GOLLUM_BASE_URL` is set; rate-limited (5 concurrent); fire-and-forget
- **ClientLogEventSchema**: Zod schema aligned with FE→BE contract

### Frontend
- **BackendLogTransport**: Uses `fetch` with `keepalive` for normal sends; flushes pending events via `sendBeacon` on `pagehide` (tab close, refresh)
- **backendLoggerProvider**: Implements `LoggerProviderConfig`; maps `log`/`reportError` to `ClientLogEvent` payloads
- **createClientEvent** / **parseExtra**: Build events and parse extras (sessionId, connectionId, etc.)
- **serializeClientEvent**: Makes payloads JSON-safe (Errors → object, circular refs → `[unserializable]`)

### libs/common
- **ClientLogEvent** type and **SerializedError** for FE/BE contract
- **LoggerBase**: Buffer early logs before `setup()`; flush on provider init

### VonageVideoClient
- Event names: `EnterMeeting`, `HandleSessionDisconnected`, `CallReconnecting`, `CallReconnected`
- Added `partnerId` (apiKey) for analytics
- Added `eventSource` in extras for context

## Testing
- Unit tests for createClientEvent, parseExtra, serializeClientEvent, BackendLogTransport, backendLoggerProvider
- Backend logger and gollumClientService tests

### Usage examples
The logging provider is already used by the Vonage video client.  
Examples can be found in:

`frontend/src/utils/VonageVideoClient/vonageVideoClient.ts`

Some emitted events include:
- `CallReconnecting`
- `CallReconnected`
- `CallStarted`
- `CallEnded`
- `reportError(...)`

These flows helped validate that events are delivered during normal operation and also when the page is refreshed or the tab is closed (sendBeacon fallback).

1. Run the project, join a meeting room, then open DevTools → Network.
2. In the Network tab, look for requests to `client-log`.
3. Confirm that an `EnteringMeeting` event is sent.
4. Copy the `sessionId` 
5. If you have Kibana access, search by the `sessionID` to verify the event is ingested.

Screenshot 2026-03-05 at 11 31 30

Resolves [VIDSOL-509](https://jira.vonage.com/browse/VIDSOL-509)

#### Checklist
[ ] Resolves a `Known Issue`.


backend/routes/index.ts
```diff
@@ -3,12 +3,14 @@ import healthRoute from './health';
...
 import wellKnownRouter from './wellKnown';
+import loggerRouter from './logger';
 
 const router = Router();
 
 router.use('/_', healthRoute);
 router.use('/session', sessionRouter);
 router.use('/feedback', feedbackRouter);
 router.use('/.well-known', wellKnownRouter);
+router.use('/internal/client-logs', loggerRouter);
```
OscarFava: NIT: do we need to add a subpath /internal/client-logs or could we directly use client-logs?

backend/routes/logger.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Backend logging endpoint. Validates the payload and forwards to Gollum/HLG when configured.
+ */
+loggerRouter.post('/', (req: Request, res: Response) => {
+  const parsed = ClientLogEventSchema.safeParse(req.body);
+
+  if (!parsed.success) {
+    return res.status(400).json({
+      message: 'Invalid log payload',
```
OscarFava: Nit: Could we add some specific error message?

backend/routes/logger.ts
```diff
@@ -0,0 +1,28 @@
...
+    });
+  }
+
+  const event = parsed.data;
+  void forwardToGollum(event).catch(console.error);
```
OscarFava: Nit: It is possible to receive any kind of error from this forwardToGollum function which we should send to the FE? I see we are always sending a 200 here.

backend/services/gollumClientService.ts
```diff
@@ -0,0 +1,77 @@
...
+    ...mapToGollumPayload(event),
+    serverReceivedTime: Date.now(),
+  };
+
+  try {
+    await limit(() =>
+      axios.post(gollumUrl, body, {
+        headers: { 'Content-Type': 'application/json' },
+        timeout: 5000,
+        validateStatus: (status) => status < 500,
```
OscarFava: What is that? Are we accepting 400 as a valid value?

OscarFava (commented):
Great work, some minor comments.

backend/routes/index.ts
Hossein-Movahed: I used /internal to signal that this is a client telemetry ingestion endpoint rather than part of the public API surface. It’s not intended to be consumed by external clients, just our web app sending logs/errors.

Happy to switch to `/client-logs` if we prefer flatter routing , I don’t have a strong preference, I just wanted to distinguish it from normal API resources.

backend/routes/logger.ts
Hossein-Movahed: I will include Zod validation details in error

backend/routes/logger.ts
Hossein-Movahed: We intentionally always return 200 here. The client’s responsibility ends once we accept the payload. Forwarding to Gollum is a backend concern, if Gollum is down, that’s an infra issue we handle with monitoring/alerting, not by surfacing 5xx to the client. Returning 500 would also encourage retries, which could cause duplicate events. forwardToGollum already catches and logs its own errors, so it doesn’t throw; the `.catch(console.error)` is just a safety net for unexpected failures.

backend/services/gollumClientService.ts
Hossein-Movahed: Good question 🙂 this is intentional.

We’re intentionally treating 4xx as non-fatal for this telemetry forwarder. 400 means Gollum rejected the payload (e.g. schema mismatch, old client), not a backend failure. Treating 4xx as success avoids log spam and keeps the logging pipeline off the critical path. The client still gets 200; we’ve accepted the log. If we need visibility into Gollum rejections, we can add metrics or sampled logging instead of treating every 4xx as an error.

OscarFava (dismissed):

backend/routes/index.ts
johnny-quesada-developer: If we don't want clients to use it we need to close the endpoint by implementing some sort of role or permissions validations (this usually is done with a token)... adding "internal" doesn't make the endpoint be internal... Better to keep it flat. 

cc: @OscarFava

backend/routes/logger.ts
```diff
@@ -0,0 +1,34 @@
...
+const loggerRouter = Router();
+
+/** JSON body parser with 50kb limit (logger route only; avoids large payloads). */
+loggerRouter.use(express.json({ limit: '50kb' }));
+
+/**
+ * Backend logging endpoint. Validates the payload and forwards to Gollum/HLG when configured.
+ */
+loggerRouter.post('/', (req: Request, res: Response) => {
+  const parsed = ClientLogEventSchema.safeParse(req.body);
```
johnny-quesada-developer, line 25: lets use a try catch statement and fail fast if there is a data violation... 
```ts
const parsed = assertResult(
  () => ClientLogEventSchema.parse(req.body),
  (error) => ({
    fallbackConfig: {
      fallbackMessage: error.message ?? 'Invalid payload',
      statusCode: StatusCode.ClientErrorBadRequest,
      severity: 'error',
    },
  })
);

```

backend/routes/logger.ts
johnny-quesada-developer: I agree with @OscarFava.

Returning just "Invalid log payload" is too generic. What exactly is invalid?

We are already using Zod with a detailed schema validation, so we have structured information about what failed and where. We should leverage that instead of collapsing everything into a generic message.

The ApplicationError should be adjusted to safely handle validation errors and expose structured validation details. The endpoint should return enough information for the client to clearly understand what happened and react accordingly.

At minimum, the client should receive:
- A proper HTTP status code (e.g. 400 or 422)
- A stable error code
- A list of validation issues with paths and messages

For example:
```json
{
  "message": "Invalid request",
  "severity": "error",
  "statusCode": 400, // Bad request
  "issues": [
    {
      "path": ["username"],
      "message": "Invalid input: expected string, received number"
    },
    {
      "path": ["favoriteNumbers", 1],
      "message": "Invalid input: expected number, received string"
    }
  ]
}
```

In this case because this are beacons we are not going to be reviewing this errors, but we need to set an standard for all endpoints, the api should be always consistent, and with the proper error json we can use zod on the client to display the errors in a more human like format **[like](https://zod.dev/error-formatting)**.

Also, you are missing something important, as I said, we are not going to review this possible exceptions cause we'll be sending [beacons](https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API) , so, probably what we want is an error handler middleware and store the violations on this endpoint for us.

backend/routes/logger.ts
johnny-quesada-developer: Hi @Hossein-Movahed , 

- The standard for logs is `204 No Content`...  
- Lets use **attemp(..** to make the indent more clear than and empty catch... 

Also, about the topic of the retries, retrying is not wrong, we should have a light retry with **idempotentCallbackWithRetry**, and if we were having something more critica like some logs that they need to be saved mandatorily we'll have to make the events **Idempotents** but Gollum does support that.

Finally, the `p-limit` is totally unnecessary, is not doing what you believe it does.. The endpoint receives one event at the time, we are not processing batches, so `const limit = pLimit(5);` you are never hitting limit with more than one request at the time... let's please remove that.

backend/services/gollumClientService.ts
```diff
@@ -0,0 +1,77 @@
+import axios from 'axios';
+import loadConfig from '../helpers/config';
+import type { ClientLogEvent } from '../types/ClientLogEventSchema';
+import pLimit from 'p-limit';
+
+// limit to 5 concurrent requests
+const limit = pLimit(5);
```
johnny-quesada-developer: This limit is not doing anything , lets remove it @Hossein-Movahed

backend/services/gollumClientService.ts
```diff
@@ -0,0 +1,77 @@
...
+    partnerId: event.partnerId,
+  };
+}
+
+/**
+ * Forwards a validated ClientLogEventSchema to Gollum/HLG.
+ * Fire-and-forget: logs errors but does not throw so the /logger route can always return 200.
+ */
+export async function forwardToGollum(event: ClientLogEvent): Promise<void> {
+  if (!gollumUrl) {
```
johnny-quesada-developer: this cannot just silently skip. what happen if we made a mistake and there is not gollumUrl?? How long it will take us to realise of it @Hossein-Movahed ?

backend/services/gollumClientService.ts
```diff
@@ -0,0 +1,77 @@
...
+  guid: string;
+  clientVersion?: string;
+  logVersion?: string;
+  userAgent?: string;
+  name?: string;
+  componentId?: string;
+  partnerId?: string;
+};
+
+function mapToGollumPayload(event: ClientLogEvent): GollumClientEventPayload {
```
johnny-quesada-developer: Why creating a new type?? **ClientLogEvent** and **GollumClientEventPayload** are basically identical.  

But also, the payload type probably can live on **libs/common**,

backend/services/gollumClientService.ts
```diff
@@ -0,0 +1,77 @@
...
+    action: event.action,
+    variation: event.variation,
+    sessionId: event.sessionId,
+    connectionId: event.connectionId,
+    clientSystemTime: event.clientSystemTime,
+    payload: event.payload ?? null,
+    source: event.source,
+    guid: event.guid,
+    clientVersion: event.clientVersion,
+    logVersion: event.logVersion ?? '2',
```
johnny-quesada-developer: why 2? @Hossein-Movahed

backend/services/gollumClientService.ts
```diff
@@ -0,0 +1,77 @@
...
+};
+
+function mapToGollumPayload(event: ClientLogEvent): GollumClientEventPayload {
+  return {
+    action: event.action,
+    variation: event.variation,
+    sessionId: event.sessionId,
+    connectionId: event.connectionId,
+    clientSystemTime: event.clientSystemTime,
+    payload: event.payload ?? null,
```
johnny-quesada-developer: better not to add default values so we can see exactly what we are sending from the client

backend/services/gollumClientService.ts
```diff
@@ -0,0 +1,77 @@
...
+    return;
+  }
+
+  const body = {
+    ...mapToGollumPayload(event),
+    serverReceivedTime: Date.now(),
+  };
+
+  try {
+    await limit(() =>
```
johnny-quesada-developer: no necessary... we are not processing batches

backend/services/gollumClientService.ts
```diff
@@ -0,0 +1,77 @@
...
+      axios.post(gollumUrl, body, {
+        headers: { 'Content-Type': 'application/json' },
+        timeout: 5000,
+        validateStatus: (status) => status < 500,
+      })
+    );
+  } catch (error: unknown) {
+    const errorMessage = error instanceof Error ? error.message : 'Unknown error occurred';
```
johnny-quesada-developer: A helper should not fail silently. If there is an error, the router should catch it.

backend/tests/logger.test.ts
```diff
@@ -0,0 +1,140 @@
+/* eslint-disable @typescript-eslint/await-thenable */
```
johnny-quesada-developer: please do not disable rules... we have disable **await-thenable** in some file to avoid a minor refactor but new code should respect the rule

backend/tests/logger.test.ts
```diff
@@ -0,0 +1,140 @@
...
+import { afterAll, beforeAll, beforeEach, describe, expect, it, jest } from '@jest/globals';
+import request from 'supertest';
+import { Server } from 'http';
+import mockOpentokConfig from '../helpers/__mocks__/config';
+
+await jest.unstable_mockModule('../helpers/config', mockOpentokConfig);
+
+const mockForwardToGollum = jest.fn<() => Promise<void>>().mockResolvedValue(undefined);
```
johnny-quesada-developer: you can mock axios instead of mocking all the helper... mocking should be minimum all the time

backend/types/ClientLogEventSchema.ts
```diff
@@ -0,0 +1,26 @@
...
+  logVersion: z.string().optional(),
+  name: z.string().optional(),
+});
+
+/**
+ * Frontend → Backend logging payload contract.
+ *
+ * **Must match** `libs/common/src/logger/ClientLogEvent.ts` so FE and BE can work in parallel.
+ */
+export type ClientLogEvent = z.infer<typeof ClientLogEventSchema>;
```
johnny-quesada-developer, line 26: this type is duplicated in **vonage-video-react-app/libs/common/src/logger/ClientLogEvent.ts**

there should be only one type... that's why we have **lib/common**

backend/types/ClientLogEvent.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const ClientLogEventSchema = z.object({
+  action: z.string().min(1),
+  variation: z.string().optional(),
+  sessionId: z.string().optional(),
+  connectionId: z.string().optional(),
+  clientSystemTime: z.number(),
+  payload: z.record(z.string(), z.unknown()).optional(),
+  source: z.string(),
+  guid: z.string().min(1),
```
johnny-quesada-developer, line 15:
```tsx
  guid: z.guid().min(1),

```

backend/types/ClientLogEvent.ts
johnny-quesada-developer, line 1: The other schemas names follow this pattern `ClientLogEvent.schema.ts` let's keep it consistent

frontend/src/logger/event/createClientEvent.ts
```diff
@@ -0,0 +1,40 @@
+import type { ClientLogEvent } from '@common/logger';
+import getAppVersion from '@utils/getAppVersion';
+
+/** Correlation id: same for all events from this page load. Enables grouping in Kibana. */
+const correlationIdForPageLoad = crypto.randomUUID();
+
+type ClientEventInput = {
```
johnny-quesada-developer, line 8: does everything needs to be optional?  that downgrades the quality of the typing

frontend/src/logger/event/parseExtras.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Extracts reserved keys (sessionId, connectionId, timestamp, partnerId) from extra.
+ * Puts remaining keys into payload. Validates types (string for ids, number for timestamp).
+ */
+export function parseExtra(extra?: Record<string, unknown>): ParsedExtra {
+  const { sessionId, connectionId, timestamp, partnerId, ...rest } = extra ?? {};
+
+  return {
+    sessionId: typeof sessionId === 'string' ? sessionId : undefined,
```
johnny-quesada-developer:
```tsx
isString(sessionId) ? sessionId : undefined
```

But which other value could have??? according to the typing is rather string or undefined, so the validation is no necessary.

frontend/src/logger/providers/backendLoggerProvider.ts
```diff
@@ -0,0 +1,63 @@
...
+import { createClientEvent, parseExtra } from '../event';
+
+export function createBackendLoggingProvider(): LoggerProviderConfig {
+  const transport = new BackendLogTransport();
+
+  return {
+    verbose: false,
+
+    log: (event: string, extra?: Record<string, unknown>) => {
+      // Fire-and-forget: logging must not throw or block the UI; errors are intentionally swallowed.
```
johnny-quesada-developer: why would the code break here in normal circumstances ??  transport.send also has a try catch... Let's please don't arbitrarily wrap everything into try catch ...  here we do not need a try catch, we need to make sure that the logic will work correctly

frontend/src/logger/providers/backendLoggerProvider.ts
```diff
@@ -0,0 +1,63 @@
...
+        transport.send(clientEvent);
+      });
+    },
+
+    reportError: (error: unknown, extra?: Record<string, unknown>) => {
+      // Fire-and-forget: logging must not throw or block the UI; errors are intentionally swallowed.
+      tryCatch(() => {
+        const parsedExtra = parseExtra(extra);
+
+        const variation =
```
johnny-quesada-developer: you can use a validation we already have `isErrorLike`

frontend/src/logger/providers/backendLoggerProvider.ts
```diff
@@ -0,0 +1,63 @@
...
+          ...parsedExtra,
+        });
+
+        transport.send(clientEvent);
+      });
+    },
+
+    reportError: (error: unknown, extra?: Record<string, unknown>) => {
+      // Fire-and-forget: logging must not throw or block the UI; errors are intentionally swallowed.
+      tryCatch(() => {
```
johnny-quesada-developer: remove the tryCatch, and for the cases where we really want just to try please use `attemp`

frontend/src/logger/providers/backendLoggerProvider.ts
```diff
@@ -0,0 +1,63 @@
...
+        const parsedExtra = parseExtra(extra);
+
+        const variation =
+          error instanceof Error
+            ? error.name
+            : typeof error === 'object' && error !== null && 'name' in error
+              ? String((error as { name: unknown }).name)
+              : 'Error';
+
+        const errorPayload = {
```
johnny-quesada-developer: same... `isErrorLike`... but Ideally the ApplicationClientError should be able to receive the error as src and export the correct format so we dont need to duplica this extraction everywhere

frontend/src/logger/providers/backendLoggerProvider.ts
```diff
@@ -0,0 +1,63 @@
...
+          error:
+            error instanceof Error
+              ? { message: error.message, name: error.name, stack: error.stack }
+              : error,
+          ...(parsedExtra.payload ?? {}),
+        };
+
+        const clientEvent = createClientEvent({
```
johnny-quesada-developer: could we just simplify this ?? why not just the **BackendLogTransport** does everything ? whats the point of separating the things here if we are preparing a type of parameter that is specific for BackendLogTransport

johnny-quesada-developer (changes requested):

frontend/src/logger/transport/backendLogTransport.ts
```diff
@@ -0,0 +1,70 @@
...
+import type { ClientLogEvent } from '@common/logger';
+import { serializeClientEvent } from './serializeClientEvent';
+
+type PendingEntry = { id: number; body: string };
+
+/**
+ * Transport that POSTs ClientLogEvent to the backend.
+ * Uses fetch for normal sends; flushes any in-flight events via sendBeacon on pagehide (tab close, refresh).
+ */
+export class BackendLogTransport {
```
johnny-quesada-developer, line 11: this is a lot more complex than what it should be... for **telemetry** which is what we are doing here... you need to send an forget, we don't care about the results... you should use **navigator.sendBeacon** and fallback to **fetch** with **keepalive** only if   **sendBeacon** is not available

frontend/src/logger/transport/serializeClientEvent.ts
```diff
@@ -0,0 +1,51 @@
+import tryCatch from '@common/execution/tryCatch';
+import isErrorLike from '@common/assertions/isErrorLike';
+import type { ClientLogEvent } from '@common/logger';
+
+type Primitive = string | number | boolean | null;
+
+function serializeValue(value: unknown): Primitive | Record<string, unknown> {
```
johnny-quesada-developer: Hi @Hossein-Movahed ,

you can use :
```ts
import { formatToStore } from 'json-storage-formatter';
```

That one will preserve a lot more of metadata... **JSON.stringify** alone could make you lose a lot of information.

backend/services/gollumClientService.ts
Hossein-Movahed: Its what we send atm with `LogOnConnect`

OscarFava (dismissed):

backend/errors/ValidationError.ts
```diff
@@ -0,0 +1,17 @@
+export type ValidationIssue = {
+  path: (string | number)[];
+  message: string;
+};
+
+export class ValidationError extends Error {
```
johnny-quesada-developer: Hi @Hossein-Movahed ,

We have **@common => ApplicationError**, we don't need another class, just extends **ApplicationError** to also carry issues and path...

backend/routes/logger.ts
```diff
@@ -0,0 +1,48 @@
...
+    message: issue.message,
+  }));
+}
+
+/**
+ * Backend logging endpoint. Validates the payload and forwards to Gollum/HLG when configured.
+ * Fails fast on data violation. Returns 204 No Content on success (standard for logs).
+ */
+loggerRouter.post('/', (req: Request, res: Response, next: express.NextFunction) => {
+  const { result: event, error } = tryCatch(() => ClientLogEventSchema.parse(req.body));
```
johnny-quesada-developer: you can use `ClientLogEventSchema.safeParse`.

johnny-quesada-developer (changes requested):

backend/routes/logger.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  if (!event) {
+    next(new Error('Unexpected: parse succeeded but result is null'));
+    return;
+  }
+  void attempt(() => forwardToGollum(event as ClientLogEvent), console.error);
+  return res.sendStatus(204);
+});
+
+loggerRouter.use(validationErrorHandler);
```
johnny-quesada-developer: We don't need an error middleware for only one endpoint, you can achieve the same with a try catch... If you want to add an error middleware should be at application level, not just for one router, and it should be a little more robust... The idea will be having something like:

```ts
const errorHandler = async (error, req, res, _next) => {
  // try to log the error...
  tryToLogApplicationError(error);

  if (isDevelopment) {
    console.error('Development error log:', error);
  }

  const applicationError = new ApplicationServerError({
    src: error,
    /**
     * If the error is not a custom ApplicationError,
     * We'll fallback to internal server error
     */
    fallbackConfig: {
      fallbackMessage: 'Internal server error',
      statusCode: StatusCodeEnum.ServerErrorInternal,
    },
  });

  const accepts = req.headers.accept || '';

  // avoids exposing sensitive information while still providing useful error information
  const safeError = applicationError.exportSafely();

  const isJsonRequest =
    accepts.includes('application/json') ||
    req.xhr ||
    req.headers?.['content-type']?.includes('application/json');

  if (isJsonRequest) {
    res.status(applicationError.statusCode).json(safeError);
    return;
  }

  const isHtmlRequest = accepts.includes('text/html');

  if (isHtmlRequest) {
    res.status(applicationError.statusCode).render('index', {
      error: safeError,
    });

    return;
  }

  // fallback to plain text response for other request types
  res.status(safeError.statusCode).send(safeError.message);
};
```

But for that you need to make sure that endpoints will never throw unhandled rejections and that everything will be catch by the error middleware, this doesn't happens on express by default..

backend/routes/logger.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (error) {
+    if (error instanceof ZodError) {
+      return next(new ValidationError(formatValidationIssues(error)));
+    }
+    next(error);
+    return;
+  }
+
+  if (!event) {
+    next(new Error('Unexpected: parse succeeded but result is null'));
```
johnny-quesada-developer: with the current **ClientLogEventSchema** the **event** will never be null without and error... This whole if will never trigger

backend/routes/logger.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Backend logging endpoint. Validates the payload and forwards to Gollum/HLG when configured.
+ * Fails fast on data violation. Returns 204 No Content on success (standard for logs).
+ */
+loggerRouter.post('/', (req: Request, res: Response, next: express.NextFunction) => {
+  const { result: event, error } = tryCatch(() => ClientLogEventSchema.parse(req.body));
+
+  if (error) {
+    if (error instanceof ZodError) {
```
johnny-quesada-developer: the error here cannot be anything but **ZodError**, the code is overcomplicated... you can just check if there is an error and if so throw an exception that will be catch by the error middleware

backend/services/gollumClientService.ts
```diff
@@ -0,0 +1,30 @@
...
+const { gollumUrl } = loadConfig();
+
+/**
+ * Forwards a validated ClientLogEvent to Gollum/HLG.
+ * Backend adds serverReceivedTime when forwarding. Gollum: POST /{source}/
+ * Throws on error so the route can catch and handle.
+ */
+export async function forwardToGollum(event: ClientLogEvent): Promise<void> {
+  if (!gollumUrl) {
+    console.warn(
```
johnny-quesada-developer: Hi @Hossein-Movahed ,

Thanks for the changes but this is not the correct pattern, miss configurations should fail loudly... otherwise imagine, we could easy spend 3 months "logging" and then realising that we didn't configure the **Gollum** url

backend/types/ClientLogEvent.ts
johnny-quesada-developer: Hi @Hossein-Movahed , the file still is call **ClientLogEvent** not ** ClientLogEvent.schema.ts**, also the folder where is located is wrong... this is not a type, this is and schema, within out project those are different kind of entities. 

Image

frontend/src/logger/event/createClientEvent.ts
```diff
@@ -0,0 +1,42 @@
...
+import getAppVersion from '@utils/getAppVersion';
+import OT from '@vonage/client-sdk-video';
+
+/** Correlation id: same for all events from this page load. Enables grouping in Kibana. */
+const correlationIdForPageLoad = crypto.randomUUID();
+
+type ClientEventInput = {
+  level: ClientLogEvent['level'];
+  action: string;
+  variation?: string;
```
johnny-quesada-developer: Hi @Hossein-Movahed,

Is variation really necessary here? What is its intended purpose? Is it a requirement coming from Gollum?

If we do need it, could we define a stricter contract or a set of expected values instead of using a plain string? Logs are much more useful when their structure is well defined.

frontend/src/logger/event/parseExtras.ts
```diff
@@ -0,0 +1,25 @@
...
+  timestamp?: number;
+  partnerId?: string;
+  payload?: Record<string, unknown>;
+};
+
+/**
+ * Extracts reserved keys (sessionId, connectionId, timestamp, partnerId) from extra.
+ * Puts remaining keys into payload. Validates types (string for ids, number for timestamp).
+ */
+export function parseExtra(extra?: Record<string, unknown>): ParsedExtra {
```
johnny-quesada-developer: why is this required? the log should probably resent to **gollum** whatever is passed down from the FE, not all the logs necessarily have sessionId.

frontend/src/logger/providers/backendLoggerProvider.ts
```diff
@@ -0,0 +1,13 @@
+import type { LoggerProviderConfig } from '@common/logger';
+import { BackendLogTransport } from '../transport';
+
+export function createBackendLoggingProvider(): LoggerProviderConfig {
```
johnny-quesada-developer: ts can infer the types based on the declared return value, so you could:

```ts
export function createBackendLoggingProvider(): LoggerProviderConfig {
  const transport = new BackendLogTransport();

  return {
    verbose: false,
    log: (...args) => transport.log(...args),
    reportError: (...args) => transport.reportError(...args),
  };
}
```

frontend/src/logger/transport/backendLogTransport.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    for (const { body } of this.pending) {
+      navigator.sendBeacon(this.endpoint, new Blob([body], { type: 'application/json' }));
+    }
+    this.pending = [];
+  };
+
+  /** Logs an event. Fire-and-forget: uses attempt so errors do not propagate. */
+  log(event: string, extra?: Record<string, unknown>): void {
+    attempt(
```
johnny-quesada-developer: **Attempt** doesn’t make sense here. Everything happening inside the attempt is controlled by us, so why would it fail?

frontend/src/logger/transport/backendLogTransport.ts
```diff
@@ -0,0 +1,127 @@
...
+      () => {
+        const parsed = parseExtra(extra);
+        const clientEvent = createClientEvent({
+          level: 'info',
+          action: event,
+          variation: 'Success',
```
johnny-quesada-developer: 1- Would be better to extend the logger to have something like logSessionEvent?  so that we can required the specific properties of the session that we want to log? 

2- are we logging **events** or **actions**?  Better to use only one term.

We a can extend the interface if it looks like we are not requesting all the necessary parameters instead of just hardcoding... we could also use method overload

frontend/src/logger/transport/backendLogTransport.ts
```diff
@@ -0,0 +1,127 @@
...
+        this.send(clientEvent);
+      },
+      (err) => console.error('[logger] reportError failed:', err)
+    );
+  }
+
+  /** Sends a ClientLogEvent using fetch with light retry; adds to pending for pagehide fallback. */
+  private send(event: ClientLogEvent): void {
+    const safeEvent = serializeClientEvent(event);
+    const body = JSON.stringify(safeEvent);
```
johnny-quesada-developer: we mention this earlier... **JSON.stringify** can and will make us lose data:

Image

frontend/src/logger/transport/backendLogTransport.ts
```diff
@@ -0,0 +1,127 @@
...
+    const id = this.nextId++;
+
+    tryCatch(() => {
+      this.pending.push({ id, body });
+
+      idempotentCallbackWithRetry(
+        () =>
+          fetch(this.endpoint, {
```
johnny-quesada-developer: If we are using axios lets use axios everywhere... but, here actually we need to use **navigator.sendBeacon**, this fetch could be necessary if we we handling something like **logger.logCriticalEvent(**.. something we know we cannot afford to lose, but for telemetry the standard is to use **navigator.sendBeacon** to avoid delaying blocking the ui.

For critical logs the solution could be even more complex implying temporal saving the logs on indexDB or something, is not what we are looking for at least not now.

frontend/src/logger/transport/backendLogTransport.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  /** Sends a ClientLogEvent using fetch with light retry; adds to pending for pagehide fallback. */
+  private send(event: ClientLogEvent): void {
+    const safeEvent = serializeClientEvent(event);
+    const body = JSON.stringify(safeEvent);
+    const id = this.nextId++;
+
+    tryCatch(() => {
+      this.pending.push({ id, body });
```
johnny-quesada-developer: We don't need to this... for critical logs we wouldn't store them on volatile memory.

frontend/src/logger/transport/backendLogTransport.ts
```diff
@@ -0,0 +1,127 @@
...
+      idempotentCallbackWithRetry(
+        () =>
+          fetch(this.endpoint, {
+            method: 'POST',
+            body,
+            keepalive: true,
+            headers: { 'Content-Type': 'application/json' },
+          }).then((res) => {
+            if (res.ok) {
+              const index = this.pending.findIndex((p) => p.id === id);
```
johnny-quesada-developer: this is no needed for simple logs...

frontend/src/logger/transport/backendLogTransport.ts
```diff
@@ -0,0 +1,127 @@
...
+          }),
+        { retries: 2, delayMs: 200 }
+      ).catch(() => {
+        // Keep in pending; sendBeacon will flush on pagehide
+      });
+    });
+  }
+
+  /** Optional: remove pagehide listener if provider is destroyed. */
+  destroy(): void {
```
johnny-quesada-developer: We wanted to implement something like **send and forget**. For that purpose, we probably should not handle a growing pending queue. This can easily grow in memory allocation and become a problem for the actual app.

frontend/src/logger/transport/serializeClientEvent.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Makes event.payload JSON-serializable: Errors → { message, name, stack }; objects/arrays → JSON string;
+ * circular refs → '[unserializable]'. Returns event unchanged if payload is absent.
+ */
+export function serializeClientEvent(event: ClientLogEvent): ClientLogEvent {
+  if (event.payload == null) {
+    if (event.payload === undefined) {
+      return event;
+    }
+    return { ...event, payload: undefined };
```
johnny-quesada-developer: you'll never get if `(event.payload == undefined) {` after checking `event.payload == null`... and whats the purpose of changing the value of the payload?  this could create a lot of confusion when reviewing logs since we'll be overriding the actual payload value.

frontend/src/logger/transport/serializeClientEvent.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  const payload: Record<string, unknown> = {};
+
+  for (const [k, v] of Object.entries(event.payload)) {
```
johnny-quesada-developer: same, as before... you can use **formatToStore** and delete this whole file..

frontend/src/logger/FrontendLogger.ts
```diff
@@ -8,9 +9,26 @@ export type FrontendLoggerProviderConfig = LoggerProviderConfig & {
 };
 
 /**
- * FrontendLogger class extends the LoggerBase class and provides a way to log errors and events to the console.
+ * FrontendLogger extends LoggerBase with React error callbacks (onCaughtError, onUncaughtError, onRecoverableError).
+ * Logs are sent via the configured provider (e.g. backend → Gollum); not to the console.
  */
 export class FrontendLogger extends LoggerBase {
```
johnny-quesada-developer, line 14: Class and file names they don't match

frontend/src/utils/VonageVideoClient/vonageVideoClient.ts
```diff
@@ -316,9 +315,10 @@ class VonageVideoClient extends EventEmitter<VonageVideoClientEvents> {
    */
   private handleSubscriptionError = (error: unknown) => {
     frontendLogger.reportError(error, {
-      source: 'vonageVideoClient.handleSubscriptionError',
+      eventSource: 'vonageVideoClient.handleSubscriptionError',
```
johnny-quesada-developer, line 331: We should probably have some sort of dictionary to avoid repeating sources:
```tsx
export enum LogEventSource { 
       SubscriptionError = 'vonageVideoClient.handleSubscriptionError', 
       Reconnecting = 'vonageVideoClient.handleReconnecting', 
       Reconnected = 'vonageVideoClient.handleReconnected', 
}
```

This are logs identifier, each one points to an specific event on the app, they cannot be duplicated in order to be able to correctly identify the logs later on.

libs/common/src/logger/ClientLogEvent.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+/** Allowed payload value types after serialization. */
+export type ClientLogPayloadValue = string | number | boolean | null | SerializedError;
+
+/**
+ * Frontend → Backend logging payload contract.
+ * Do not add required fields or change field types without coordinating FE and BE.
+ */
+export type ClientLogEvent = {
```
johnny-quesada-developer: if this is the contract it should be asserted... you already have the schema with zod, so you can just infer the type:

```tsx
export type ClientLogEvent = z.infer<typeof ClientLogEventSchema>;
```

You dont need this file

libs/common/src/logger/Logger.ts
```diff
@@ -23,6 +23,13 @@ export const ColorsPerFeature = {
 /** Max retries when provider resolution returns null, to avoid infinite recursion. */
 const MAX_PROVIDER_RESOLVE_RETRIES = 2;
 
+/**
+ * Events buffered when provider is not yet set. Flushed to the provider when setup() completes.
+ */
+type QueuedLog = { type: 'log'; args: [string, Record<string, unknown>] };
```
johnny-quesada-developer: If we call navigator.sendBeacon many times, the browser already handles best-effort scheduling/queuing of those requests, so we don't need to maintain our own pending queue for simple telemetry. And if we actually needed guaranteed delivery, storing the logs/queue in ephemeral memory wouldn't be reliable anyway.

libs/common/src/logger/Logger.ts
```diff
@@ -213,34 +248,59 @@ export class LoggerBase implements LoggerProviderConfig {
     }) as Omit<LoggerBase, 'group' | 'setup'>;
   }
 
+  /**
+   * Enriches context with a timestamp. Use for all provider/queue payloads so log, reportError,
+   * and any future methods attach a consistent client timestamp.
+   */
+  protected withTimestamp<T extends Record<string, unknown>>(data: T): T & { timestamp: number } {
```
johnny-quesada-developer: this is wrong, what happens if i already have a timestamp on data? this should be totally apart from the generic context... that will also prevent us from having to copy the object.

libs/common/src/logger/Logger.ts
```diff
@@ -213,34 +248,59 @@ export class LoggerBase implements LoggerProviderConfig {
...
+          ...error,
+          message: error.message,
+          name: error.name,
+          stack: error.stack,
+        }
+      : error;
+
+    const extraRecord = this.withTimestamp(extra ?? {});
+
+    if (this.provider === null) {
```
johnny-quesada-developer: this queue does the same as the subsequent promise... we are duplicating efforts aren't we?

libs/common/src/logger/Logger.ts
```diff
@@ -213,34 +248,59 @@ export class LoggerBase implements LoggerProviderConfig {
...
+   * Before setup(): event is buffered and flushed when setup() completes.
    * @param event The name of the event to log.
    * @param extra Additional data associated with the event.
    */
-  public log(event: string, extra?: Record<string, unknown>) {
-    const payload = { ...(extra ?? {}), timestamp: Date.now() };
+  public log(event: string, extra: Record<string, unknown> = {}) {
+    const payload = this.withTimestamp(extra);
+
+    if (this.provider === null) {
```
johnny-quesada-developer: Same as before: in this case, the promise chain already behaves like a pending queue... calls are executed once the provider is ready. We could simplify this by only waiting for ready and then using the logger feature directly. Right now, maintaining a separate queue looks like duplicated logic, doesn't it?

johnny-quesada-developer (changes requested):

frontend/src/logger/transport/serializeClientEvent.spec.ts
johnny-quesada-developer, line 1: This helper is probably not necessary and the test is too exhaustive for a **JSON.stringify** wrapper.

frontend/src/logger/providers/backendLoggerProvider.ts
johnny-quesada-developer: why is an extra layer of abstraction here? why not just having a **new BackendLoggengProvider**?

frontend/src/logger/transport/backendLogTransport.ts
Hossein-Movahed: @johnny-quesada-developer, I used fetch + pending queue + `pagehide` flush because I wanted retries and to re-send events that might be dropped when the tab closes. If we switch to **sendBeacon-first**, we lose retries and the `pagehide` flush becomes redundant since `sendBeacon` already handles unload. Is that the intended tradeoff for **telemetry—pure** **fire-and-forget** with no retries, if so I can simplify this and just use the `seandBeacon` as you suggested!

frontend/src/logger/transport/backendLogTransport.ts
johnny-quesada-developer: same as [here](https://github.com/Vonage/vonage-video-react-app/pull/366#discussion_r2849333862)

frontend/src/logger/transport/backendLogTransport.ts
```diff
@@ -0,0 +1,127 @@
...
+        });
+        this.send(clientEvent);
+      },
+      (err) => console.error('[logger] log failed:', err)
+    );
+  }
+
+  /** Reports an error. Fire-and-forget: uses attempt so errors do not propagate. */
+  reportError(error: unknown, extra?: Record<string, unknown>): void {
+    attempt(
```
johnny-quesada-developer: same as [here](https://github.com/Vonage/vonage-video-react-app/pull/366#discussion_r2849333862)

johnny-quesada-developer (changes requested):

frontend/src/logger/transport/backendLogTransport.ts
```diff
@@ -0,0 +1,127 @@
...
+    );
+  }
+
+  /** Sends a ClientLogEvent using fetch with light retry; adds to pending for pagehide fallback. */
+  private send(event: ClientLogEvent): void {
+    const safeEvent = serializeClientEvent(event);
+    const body = JSON.stringify(safeEvent);
+    const id = this.nextId++;
+
+    tryCatch(() => {
```
johnny-quesada-developer: `navigator.sendBeacon`

backend/errors/ValidationError.ts
Hossein-Movahed: Ah this is the new one, sure will use that instead

frontend/src/logger/transport/backendLogTransport.ts
johnny-quesada-developer: Hi @Hossein-Movahed,

The problem I see with your implementation is that it looks overcomplicated for simple, **non-critical analytics/telemetry.**

You have 3 retries when sending the fetch, but also a queue. In normal circumstances, realistically speaking, how many requests will actually reach the queue? And then the queue does not persist, and a page unload is not reliable...

Finally, you have fetch retries and a queue when the 3 retries fail, but I don’t see any **idempotencyId**. What about the risk of duplicating logs? With the current setup, you try to fix an edge case and end up creating a common pattern for duplicates and memory allocation issues.

johnny-quesada-developer (changes requested):

libs/common/src/logger/Logger.ts
johnny-quesada-developer: This structure is wrong: mandatory/expected properties must be separate from the **extra** non-validated/unstructured data.

```ts

// session event
{
   // mandatory props
   sessionId,
   timespam,
   connectionId,

  // extra: { // non-structured }
}
```

johnny-quesada-developer (changes requested):

frontend/src/logger/transport/backendLogTransport.ts
Hossein-Movahed: Good point. Adding logSessionEvent would require extending the logger API (LoggerBase, LoggerProviderConfig, BackendLogTransport, etc.) and updating callers, which feels out of scope for this PR. Let’s open a separate ticket for it and keep this PR as is.

frontend/src/logger/transport/backendLogTransport.ts
Hossein-Movahed: We can consider a `logCriticalEvent` + `IndexedDB` approach in a future if needed.

frontend/src/utils/VonageVideoClient/vonageVideoClient.ts
Hossein-Movahed: The first argument to `log`/`reportError` is the event/action (e.g. `CallReconnecting`, `CallEnded`, `EnterMeeting`). The `eventSource` in extra is a separate field (as an extra field in payload) that identifies where the log originated (e.g. `vonageVideoClient.handleReconnecting`).

If we used an enum like `LogEventSource.Reconnecting` for both, we’d end up with event and eventSource both set to `vonageVideoClient.handleReconnecting`, which would duplicate the same value. That’s why we keep `eventSource` as a distinct field with its own identifiers.

libs/common/src/logger/Logger.ts
Hossein-Movahed: The `withTimestamp` is removed as suggested

OscarFava (dismissed):

Hossein-Movahed:
vcr:deploy

Hossein-Movahed:
vcr:deploy

johnny-quesada-developer (dismissed):
LGTM, didn't test it locally.

OscarFava (dismissed):

johnny-quesada-developer (dismissed):
lgtm

rserebrennykov (dismissed):
LGTM

OscarFava (approved):

rserebrennykov (approved):
lgtm

johnny-quesada-developer (approved):

PR #369, opened 2026-02-14
VIDSOL-489: Common lib refinements

johnny-quesada-developer:
#### What is this PR doing?
The project mixes environment-specific files (Node and web). The problem is that when we try to reference common resources from other projects, it becomes complex because we have to include files while also manually excluding conflicting ones that will break in a given environment.

The solution is to keep common as it was intended: it should contain only resources shared across projects in the monorepo. Environment-specific files should live in dedicated paths.

Resolves [VIDSOL-489](https://jira.vonage.com/browse/VIDSOL-489)

#### Checklist
[ ] Resolves a `Known Issue`.


OscarFava (approved):

Hossein-Movahed (approved):
🔥

OscarFava:
QA okay.

PR #370, opened 2026-02-15
VIDSOL-492: Vonage-Handler-Core

johnny-quesada-developer:
#### What is this PR doing?
Pull request overview
This PR introduces a new backend-oriented orchestration library (libs/api) and expands libs/common with Node-specific schemas/types (Zod) intended to support Vonage Video API handler validation and orchestration.

Changes:

Added libs/common/srcNode Zod schemas/types for SessionId and archive/session option payload validation.
Added initial libs/api core/types/schemas/routing scaffolding for a video orchestrator + action payloads.
Renamed sessionId assertion (assertVonageSessionId → assertSessionId) and introduced new common types (SessionId, SameKeys) plus assertString.

General QA.

Resolves [VIDSOL-492](https://jira.vonage.com/browse/VIDSOL-492)

#### Checklist
[ ] Resolves a `Known Issue`.


libs/api/src/handlers/startArchive.ts
```diff
@@ -0,0 +1,24 @@
...
+    const { sessionId, archiveOptions } = payload;
+
+    const archive = await assertResult(
+      () => this.video$.startArchive(sessionId, archiveOptions),
+      makeThirdPartyErrorHandler('Failed to start archive')
+    );
+
+    return archive;
+  } catch (error: unknown) {
+    throw makeInternalErrorHandler('Failed to start captions')(error);
```
OscarFava: Failed to start captions -> Failed to start archive

libs/core/src/services/videoClient/createVideoClient.ts
```diff
@@ -0,0 +1,64 @@
+// eslint-disable-next-line @nx/enforce-module-boundaries
```
OscarFava, line 1: Silencing the rule is ok? Could we change IVideoRouter to be exported from the public api index.ts and accessed by alias (@api-lib), not by internal relative path?

libs/api/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './routing';
```
OscarFava, line 2: Should not we export * from './core';?

OscarFava: to fix that? // eslint-disable-next-line @nx/enforce-module-boundaries
import type { IVideoRouter } from '../../../../api/src/types/IVideoRouter';

libs/api/src/handlers/searchArchives.ts
```diff
@@ -0,0 +1,20 @@
+import { makeThirdPartyErrorHandler } from '@api-lib/errors';
+import makeInternalErrorHandler from '@api-lib/errors/handlers/makeInternalErrorHandler/makeInternalErrorHandler';
```
OscarFava: Can we fix deep import?

libs/api/src/handlers/stopArchive.ts
```diff
@@ -0,0 +1,23 @@
+import { makeThirdPartyErrorHandler } from '@api-lib/errors';
+import makeInternalErrorHandler from '@api-lib/errors/handlers/makeInternalErrorHandler/makeInternalErrorHandler';
```
OscarFava: Can we fix deep import?

libs/core/src/services/videoClient/createVideoClient.ts
```diff
@@ -0,0 +1,64 @@
...
+import type { IVideoRouter } from '../../../../api/src/types/IVideoRouter';
+import { createTRPCClient, httpBatchLink } from '@trpc/client';
+import { Prettify } from '@common/types';
```
OscarFava: Can we do import type { Prettify } from '@common/types'; ?

libs/api/src/handlers/createSession.ts
```diff
@@ -0,0 +1,26 @@
...
+        mediaMode: MediaMode.ROUTED,
+        ...sessionOptions,
+      });
+    }, makeThirdPartyErrorHandler('Failed to create session'));
+
+    const session = decodeSessionId(sessionId);
+
+    return { ...session, sessionId };
+  } catch (error) {
+    throw makeInternalErrorHandler('Failed to create session')(error);
```
OscarFava, line 22: In all handlers, the external catch rewraps with makeInternalErrorHandler, which means that a ThirdPartyError (502 Bad Gateway) becomes an InternalServerError (500) when propagated.
WDYT?

libs/api/src/handlers/createEphemeralToken.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Creates an ephemeral client token with a default 30 seconds expiration time
+ * This token is meant to be used for server-to-server operations that require a token
+ */
+function createEphemeralToken(this: IVideoClient, payload: CreateEphemeralTokenPayload): string {
+  try {
+    const { sessionId, clientTokenOptions } = payload;
+
+    const tokenOptions: ClientTokenOptions = {
+      role: clientTokenOptions?.role ?? TokenRole.MODERATOR,
```
OscarFava: Is not clientTokenOptions?.role ?? TokenRole.MODERATOR redundant? Since we are ...clientTokenOptions.

Same in expireTime

libs/api/src/routing/videoRouter/createVideoRouter.ts
```diff
@@ -0,0 +1,186 @@
...
+    return input.query(async (opts) => {
+      const payload = parser(opts.input) as Parameters<VideoClient[ActionKey]>[0];
+      const orchestrator = opts.ctx[videoRouterContext].orchestrator;
+
+      return callback(orchestrator, payload);
+    });
+  }
+
+  // prettify is necessary to hide the internal TRPC types and prevent d.ts errors.
+  return Object.assign(router as Prettify<typeof router>, {
```
OscarFava, line 170: What is that? Is a TODO?

OscarFava: Or only to prevent d.ts errors? The code seems a little weird

libs/api/src/routing/videoHandler/createVideoHandler.test.ts
```diff
@@ -0,0 +1,297 @@
...
+          apiSecret: 'test-secret',
+        },
+        videoParams: {},
+        handlersConfig: {},
+      };
+
+      expect(() => createVideoHandler(validConfig)).not.toThrow();
+    });
+
+    it('should accept Auth instance as auth config', () => {
```
OscarFava: Can we directly remove that?

libs/api/src/handlers/enableCaptions.ts
```diff
@@ -0,0 +1,36 @@
...
+    const clientToken = this.createEphemeralToken({ sessionId });
+
+    const { error } = await tryCatch(() =>
+      this.video$.enableCaptions(sessionId, clientToken, captionOptions)
+    );
+
+    if (error && isErrorLike(error)) {
+      const isLiveCaptionsAlreadyStarted = error.message
+        .toLowerCase()
+        .includes('live captions have already started for this session');
```
OscarFava, line 20: Is not this too flaky? Maybe reduce the includes text length?

libs/api/src/core/VideoClient/VideoClient.ts
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Forces VideoClient to have a method for each VideoAction
+ * and correctly types the payload and return type
+ */
+type IVideoClient = {
+  [key in VideoAction]: (this: VideoClient, ...args: Any[]) => unknown;
+};
+
+class VideoClient implements IVideoClient {
+  public readonly auth$: Auth;
```
OscarFava: In the codebase, $ is used as a convention for react-global-state-hooks stores. Could using it in class properties create semantic confusion?

OscarFava (commented):
Awesome work! I leave you some initial comments before reviewing it all.

frontend/src/Context/SessionProvider/session.tsx
```diff
@@ -306,6 +305,8 @@ const SessionProvider = ({ children, initialValue = {} }: SessionProviderProps):
   };
 
   const handleArchiveStarted = (id: string) => {
+    console.log('Archive started with ID:', id);
```
OscarFava: Can we get rid of this?

backend/server.ts
```diff
@@ -30,6 +31,20 @@ app.use(bodyParser.json());
 app.set('trust proxy', true);
 app.use(router);
 
+app.use(
+  '/v2',
+  createVideoHandler({
+    auth: {
```
OscarFava, line 38: Is not this the same we use in session.ts — makeVideoClient()? Can we unify the content?
```
function makeVideoClient() {
  return createVideoClient({
    auth: {
      authType: 'jwt',
      applicationId: process.env.VONAGE_APP_ID!,
      privateKey: process.env.VONAGE_PRIVATE_KEY!,
    },
    videoParams: {
      videoHost: process.env.VONAGE_VIDEO_HOST,
    },
  });
}
```

libs/common/srcNode/assertions/assertSessionId/assertSessionId.ts
```diff
@@ -0,0 +1,8 @@
+import { SessionId } from '../../types';
+import decodeSessionId from '../../helpers/decodeSessionId';
+
+function assertSessionId(value: unknown): asserts value is SessionId {
+  decodeSessionId(value as string);
```
OscarFava: Should we use assertString instead of casting?

libs/core/src/services/videoClient/createVideoClient.ts
johnny-quesada-developer: In this case it is better to disable the rule, since this import should not happen anywhere else. No other file in the project should import this type from the backend, which is why making an exception here is preferable. We need the type here to correctly infer the available procedures on the router and keep everything in sync.

libs/api/src/index.ts
johnny-quesada-developer: [comment](https://github.com/Vonage/vonage-video-react-app/pull/370#discussion_r2928432333)

libs/api/src/routing/videoRouter/createVideoRouter.ts
johnny-quesada-developer: Both, the as Prettify<typeof router> is to prevent errors, by flattering the type we avoid exporting private names which cause ts errors.

the other functions are going to be extension methods but now there are only place holders... I'm working on them in another branch. Is going to look like this actually:

```tsx
return Object.assign(router as TRPCRouter, {
    /**
     * Use this callback if you need to transform the raw input before it's evaluated by the handlers,
     * For example adding extra properties to the input
     */
    transform$<
      ActionKey extends PublicActionKey,
      Input = Prettify<Parameters<TRPCRouter[ActionKey]>[0]>,
    >(actionKey: ActionKey, transform: (input: unknown) => Input | Promise<Input>) {
      ...
    },

    /**
     * Use this callback to completely override the handler of a specific action.
     */
    override$<
      ActionKey extends PublicActionKey,
      Result = Prettify<ReturnType<TRPCRouter[ActionKey]>>,
    >(actionKey: ActionKey, handler: (opts: ProcedureResolverOptions<Result>) => Result) {
      ...
    },

    /**
     * Use this callback to run custom code before the original handler, for example to implement custom authorization logic.
     */
    use$<ActionKey extends PublicActionKey, Result = Prettify<ReturnType<TRPCRouter[ActionKey]>>>(
      actionKey: ActionKey,
      handler: (opts: CustomMiddlewareParameters<Result>) => NextResult | Promise<NextResult>
    ) {
      ...
    },
  });
```

where `TRPCRouter = Prettify<typeof router>`

libs/api/src/handlers/enableCaptions.ts
johnny-quesada-developer: This message appears like that in the Vonage API docs, and it can also be found when inspecting the npm package. I would prefer to be able to check if the captions of a session are available directly against Vonage, but that is not possible. So this is a workaround due to limitations in the SDK.

libs/api/src/core/VideoClient/VideoClient.ts
johnny-quesada-developer: I originally wanted to use $ as an indicator for special variables, but you're right. Let's use **_auth** instead, which is more accurate in this case. Consumers can still see the value if needed, but it should be treated more like a private property.

libs/api/src/handlers/startArchive.ts
johnny-quesada-developer: done

libs/api/src/handlers/searchArchives.ts
johnny-quesada-developer: done

libs/api/src/handlers/stopArchive.ts
johnny-quesada-developer: done

libs/core/src/services/videoClient/createVideoClient.ts
johnny-quesada-developer: Yes, it is `@common/types`

libs/api/src/handlers/createSession.ts
johnny-quesada-developer: Hi, no. These handlers check whether the error is a custom ApplicationError. If so, everything is preserved (statusCode, message, stack, etc.). Only unhandled errors (native errors) are wrapped in a generic internal server error to avoid exposing sensitive information.

libs/api/src/handlers/createEphemeralToken.ts
johnny-quesada-developer: You are right, fixed

frontend/src/Context/SessionProvider/session.tsx
johnny-quesada-developer: ill remove it.

backend/server.ts
johnny-quesada-developer: Hi, it's not the same. The **VideoClient** is the orchestrator that both the router and the handler use under the hood.

```tsx
/**
 * Creates a video handler:
 * The handler is a an express router with all the video capabilities required by Vera
 */
function createVideoHandler<
  TContext extends {
    [videoRouterContext]: {
      orchestrator: VideoClient;
    };
  },
  TMeta extends object,
  Handlers extends Partial<HandlersConfig>,
  TConfig extends VideoRouterConfig<TContext, TMeta, Handlers> &
    Omit<Parameters<typeof createExpressMiddleware>[0], 'router'>,
>({ auth, videoParams, routerOptions, handlersConfig, ...routerConfig }: TConfig) {
  return createExpressMiddleware({
    router: createVideoRouter({ auth, videoParams, routerOptions, handlersConfig }),

    createContext: () => ({}) as TContext,

    ...routerConfig,
  });
}
```

The ** makeVideoClient** creates something similar to the **VideoService** we currently have (one for opentok and one for vonage), but with more robust validation and better error handling.

**We have three levels:**

1. VideoClient – a raw TypeScript service that can be used inside any handler.
2. VideoRouter – a tRPC router that exposes the public API of the VideoClient.
3. createVideoHandler – creates an Express middleware so it can be mounted in an Express app. 

Example:
```tsx
// This is how a base implementation of video app self hosted looks on the BE
app.use('/video', createVideoClient({
  auth: {
    authType: 'jwt',
    applicationId: process.env.VONAGE_APP_ID!,
    privateKey: process.env.VONAGE_PRIVATE_KEY!,
  },
}));

// And as we mentioned in previous syncs, you could enhance this with middlewares at wild:
app.use('/video', auth, logs, createVideoClient({
  auth: {
    authType: 'jwt',
    applicationId: process.env.VONAGE_APP_ID!,
    privateKey: process.env.VONAGE_PRIVATE_KEY!,
  },
}););
```

libs/api/src/handlers/enableCaptions.ts
OscarFava: No problem! thanks

OscarFava (dismissed):

Hossein-Movahed:
I'm testing with OpenTok credentials in my `.env` and hitting a 502 on GET `/session/:room`. Looking at the new `makeVideoClient()` in `session.ts`, it hardcodes Vonage JWT auth regardless of `VIDEO_SERVICE_PROVIDER`, so anyone on the OpenTok setup will silently break.
Is dropping OpenTok support intentional as part of this migration? If so, it should be called out explicitly (add a deprecation notice). If not, we need an OpenTok fallback in the session route, the existing `videoService` already handles it for archiving/captions, so the same pattern could be applied here.

Hossein-Movahed (changes requested):
https://github.com/Vonage/vonage-video-react-app/pull/370#issuecomment-4054243908

libs/api/src/types/ApplicationRequest.ts
```diff
@@ -0,0 +1,13 @@
+import type { Request as ExpressRequest } from 'express';
+import ParamsDictionary from './ParamsDictionary';
+import type { Any } from '@common/types';
+
+export type ApplicationRequest<
+  P = ParamsDictionary,
+  ResBody = Any,
+  ReqBody = Any,
+  ReqQuery = qs.ParsedQs,
```
Hossein-Movahed: `qs.ParsedQs` is referenced in the default generic but `qs` is never imported in this file. It works today only because `qs` is a transitive dependency via `Express`, but this is fragile.
Either add `import type { ParsedQs } from 'qs'` or use the equivalent from express-serve-static-core (`import type { ParsedQs } from 'express-serve-static-core'`).

libs/api/src/routing/videoHandler/createVideoHandler.ts
```diff
@@ -0,0 +1,32 @@
+import { createExpressMiddleware } from '@trpc/server/adapters/express';
+import type { videoRouterContext } from '@api-lib/constants';
+import type { VideoClient } from '@api-lib/core';
+import type { VideoRouterConfig } from '@api-lib/schemas';
+import createVideoRouter from '../videoRouter';
+import { HandlersConfig } from '@api-lib/types/HandlersConfig';
```
Hossein-Movahed: Should go through the barrel export: `import { HandlersConfig } from '@api-lib/types'`

libs/api/src/handlers/enableCaptions.ts
```diff
@@ -0,0 +1,36 @@
+import { makeInternalErrorHandler, makeThirdPartyErrorHandler } from '@api-lib/errors/handlers';
```
Hossein-Movahed: Should use the public barrel: `import { ... } from '@api-lib/errors'` Same pattern as the other handlers (e.g. `createSession.ts`,` startArchive.ts`) , let's keep it consistent.

libs/api/src/handlers/disableCaptions.ts
```diff
@@ -0,0 +1,30 @@
+import { makeInternalErrorHandler, makeThirdPartyErrorHandler } from '@api-lib/errors/handlers';
```
Hossein-Movahed: [Same import suggestion](https://github.com/Vonage/vonage-video-react-app/pull/370/changes#r2946659566)

libs/api/src/routing/videoRouter/createVideoRouter.ts
```diff
@@ -0,0 +1,186 @@
...
+    before(_middleware: Any) {},
+
+    after(_middleware: Any) {},
+
+    override(_method: string, _path: string, _handler: Any) {},
```
Hossein-Movahed: `before`, `after`, and `override` are part of the returned public API but their bodies are empty. callers get no response and no error, so misuse is completely silent. If these aren't implemented yet, they should throw (e.g. `throw new Error('Not implemented')`) so it's obvious at runtime. If they're intentionally a `no-op`, a comment explaining why would help.

libs/api/src/types/IVideoClient.ts
```diff
@@ -0,0 +1 @@
+export type IVideoClient = import('../core/VideoClient').default;
```
Hossein-Movahed, line 1: The I-prefix on both types conflicts with our naming convention, from [copilot-instructions.md](.github/copilot-instructions.md): "Do not use abbreviations or acronyms in names" (the I prefix is considered an acronym/Hungarian notation). Prefer VideoClientApi / VideoRouterApi, or simply drop the prefix since the files already live in a types/ directory.

libs/api/src/types/IVideoRouter.ts
```diff
@@ -0,0 +1 @@
+export type IVideoRouter = import('../routing/videoRouter/createVideoRouter').IVideoRouter;
```
Hossein-Movahed, line 1: [Same naming convention](https://github.com/Vonage/vonage-video-react-app/pull/370/changes#r2946687334)

johnny-quesada-developer:
> I'm testing with OpenTok credentials in my `.env` and hitting a 502 on GET `/session/:room`. Looking at the new `makeVideoClient()` in `session.ts`, it hardcodes Vonage JWT auth regardless of `VIDEO_SERVICE_PROVIDER`, so anyone on the OpenTok setup will silently break. Is dropping OpenTok support intentional as part of this migration? If so, it should be called out explicitly (add a deprecation notice). If not, we need an OpenTok fallback in the session route, the existing `videoService` already handles it for archiving/captions, so the same pattern could be applied here.

you have wrong credentials as we spoke... new development goes specifically and exclusively with vonage provider, so you need to update your .env

libs/api/src/types/IVideoClient.ts
johnny-quesada-developer: This is not against the instructions. We need this distinction to indicate that this refers only to the type and not the actual class. This is a strategy used to avoid issues with circular dependencies.

libs/api/src/types/IVideoRouter.ts
johnny-quesada-developer: This is not against the instructions. We need this distinction to indicate that this refers only to the type and not the actual class. This is a strategy used to avoid issues with circular dependencies.

OscarFava (dismissed):

johnny-quesada-developer:
vcr:deploy

libs/api/src/handlers/enableCaptions.ts
johnny-quesada-developer: done

libs/api/src/routing/videoHandler/createVideoHandler.ts
johnny-quesada-developer: done

libs/api/src/types/ApplicationRequest.ts
johnny-quesada-developer: done

OscarFava (approved):

libs/api/src/routing/videoRouter/createVideoRouter.ts
johnny-quesada-developer: done

libs/api/src/handlers/disableCaptions.ts
johnny-quesada-developer: done

OscarFava (approved):

johnny-quesada-developer:
vcr:deploy

Hossein-Movahed (approved):

johnny-quesada-developer:
vcr:deploy

PR #396, opened 2026-03-03
VIDSOL-623: Video Stats Overlay in Waiting Room

czoli1976:
#### What is this PR doing?

Adds a semi-transparent video stats badge to the upper-left corner of the waiting room video preview tile, displaying the camera's current resolution and frame rate (e.g. `720p 30fps`).

This was inspired by the UI pattern used in [Riverside.fm](https://riverside.fm) and other professional video tools — it feels like a genuinely useful little touch that lets the user immediately know what resolution and frame rate their camera will be sending *before* they even join the call, without any guesswork.

**Architecture overview:**
- New `useVideoStats` hook polls the Vonage publisher every second for resolution (`videoWidth()` / `videoHeight()`) and frame rate (via `MediaStreamTrack.getSettings()`). Stats are also read immediately when the publisher instance changes, so the UI never shows stale values while waiting for the first poll tick.
- New `VideoStatsOverlay` component renders the formatted badge. Positioning is owned by the parent (`VideoContainer`) per component guidelines. Shows `–` as a fallback when stats are not yet available.
- The feature is fully opt-in, controlled by a new `SHOW_VIDEO_STATS` environment variable (default: `false`), consistent with all other feature flags in `frontend/src/env.ts`.
- Full unit test coverage for both the hook and the component.

**Review feedback addressed:**
- Fixed stale stats when publisher instance changes: stats are now updated during render via the derived-state pattern rather than waiting for the next poll tick.
- Fixed empty dark pill when stats are unavailable: badge always shows `–` as a fallback.
- Renamed `fps` identifiers to `frameRate`/`frameRateLabel` to comply with the no-acronyms guideline.
- Added `SHOW_VIDEO_STATS` to Vite's explicit `appEnvKeys` list so it is correctly injected at build time.
- Improved SDK-specific inline and JSDoc comments in `useVideoStats`: explains why `getVideoSource().track.getSettings().frameRate` is used (preview publisher is not in a session, making `session.getPublisherStats()` unavailable), and why `videoWidth()`/`videoHeight()` are normalised from `undefined` to `null`.

#### How should this be manually tested?

1. In `vcrBuild.env.sh`, add/set `export SHOW_VIDEO_STATS=true` (or pass it as an env var when starting the dev server)
2. Start the app and navigate to the waiting room
3. Grant camera access — a small dark badge (e.g. **`720p 30fps`**) should appear in the top-left corner of the video preview tile
4. The frame rate value updates live every second
5. Toggle your camera off — the badge should disappear
[...]

PR #402, opened 2026-03-07
VIDSOL-623: docs: add coding rules to Copilot instructions - hook patterns and American English

czoli1976:
## What and Why

This PR adds three new rules to `.github/copilot-instructions.md` based on mistakes caught during code review and CI failures in PR #396 (VIDSOL-623).

#### 1. American English spelling
The project uses `cspell` configured for American English. British spellings such as *synchronised*, *initialise*, *behaviour*, or *cancelled* fail the `👮 QA` CI job. This rule makes the constraint explicit so contributors know before they push.

#### 2. No `setState` directly in effect bodies
The project enforces a non-standard `react-hooks/set-state-in-effect` ESLint rule that flags any synchronous `setState` call at the top level of a `useEffect`/`useLayoutEffect` callback. The fix is to wrap the call inside a **named helper function** declared within the effect. This rule documents the pattern with a before/after example.

#### 3. No `setState` during render (derived-state pattern)
Calling `setState` inside the render body when a prop differs from a tracked state value can cause infinite render loops if the caller passes a non-stable reference (e.g. an object literal). The correct approach is to use a `useRef` to track the last-seen prop and derive the return value directly during render — ref mutation is safe during render and does not trigger a re-render. This rule documents the anti-pattern and the correct ref-derived-return implementation.

---

Reviewers: please pay particular attention to whether the code examples are clear and actionable. The goal is for Copilot (and human contributors) to catch these patterns before code review.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1684,13 +1684,13 @@ const Component = () => {
 
 ## setState in effects and during render
 
-- **Rule:** Never call `setState` directly inside an effect body (`useEffect` or `useLayoutEffect`). The project enforces `react-hooks/set-state-in-effect`, which flags any synchronous `setState` call at the top level of an effect callback.
-- **Rule:** When you need to update state from inside an effect, wrap the call in a **named helper function** declared inside the effect. Calls through a function are not flagged by the rule.
+- **Rule:** Never call `setState` directly inside an effect body (`useEffect` or `useLayoutEffect`). To follow React best practices, enable the ESLint rule `react-hooks/no-direct-set-state-in-use-effect`, which flags any synchronous `setState` call at the top level of an effect callback.
+- **Rule:** When you need to update state from inside an effect, wrap the call in a **named helper function** declared inside the effect instead of calling `setState` inline.
 
 **Violation:**
 
 ```tsx
-// Bad: setState called directly in effect body — triggers react-hooks/set-state-in-effect
+// Bad: setState called directly in effect body — triggers react-hooks/no-direct-set-state-in-use-effect
 useEffect(() => {
     setStats(readStats(publisher)); // ❌ direct setState in effect
 }, [publisher]);
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1679,7 +1679,6 @@ const Component = () => {
         ...
     );
 };
-```
 ---
 
 ## setState in effects and during render
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1776,3 +1776,21 @@ const useVideoStats = (publisher: Publisher | null) => {
     return stats;
 };
 ```
+
+---
+
+# CI Validation Workflow
+
+These rules define how to validate changes before and after pushing, to avoid wasting CI cycles on issues that can be caught locally.
+
+- **Rule:** Before pushing a branch, always run the linter and tests locally for affected projects. Only push once local checks pass.
+
+```bash
+# Run lint and tests only for projects affected by your changes
+npx nx affected --target=lint
+npx nx affected --target=test
+```
+
+- **Rule:** After pushing, check the CI workflow status. Do not mark a task complete while any CI job is pending or failing.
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -54,6 +54,14 @@ TypeScript version: `^5.8.3`
 - **Rule:** Do not add new state management libraries. Use only existing tooling.
 - **Rule:** Components must be kept small, focused, and composable.
 
+## Spelling
```
johnny-quesada-developer, line 57: I'd prefer to rely on the plugin doing its job rather than unnecessarily increasing the docs. Don't you think this is redundant, @czoli1976? Also, if these instructions are specifically for AI models, I'd argue that it is unlikely for them to have to many grammar mistakes...

.github/copilot-instructions.md
```diff
@@ -1671,4 +1679,117 @@ const Component = () => {
         ...
     );
 };
-```
\ No newline at end of file
+---
```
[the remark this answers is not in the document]
johnny-quesada-developer: Hi @czoli1976, it looks like the code example got cut off here.

.github/copilot-instructions.md
```diff
@@ -1671,4 +1679,117 @@ const Component = () => {
...
+
+    syncStats();                              // called through a function ✓
+    const id = setInterval(syncStats, 1000);
+    return () => clearInterval(id);
+}, [publisher]);
+```
+
+---
+
+- **Rule:** Never call `setState` during the render function body (derived-state-during-render pattern). This can cause render loops when a caller passes a non-stable reference, and React may log warnings.
```
johnny-quesada-developer, line 1722: Hi @czoli1976,

I’m not sure these rules are beneficial.

They implicitly encourage relying on **useEffect** as a **side-effect listener**, which tends to push code toward an **observable-style pattern**. In practice, this often makes components harder to follow, maintain, and debug.

Whenever possible, state transitions should happen in a linear flow during render or in event handlers, rather than being triggered indirectly through effects.

In general, effects should be used to hook into the component lifecycle or synchronize with external systems, not as listeners or triggers for internal state changes.

We already have some rules addressing this area, and we also have the derived-state-during-render rule enabled. Because of that, adding this rule seems both redundant and potentially misleading.

Given this, I think it would be better not to introduce these rules:
- **Rule:** Never call `setState` directly inside an effect body (`useEffect` or `useLayoutEffect`). To follow React best practices, enable the ESLint rule `react-hooks/no-direct-set-state-in-use-effect`, which flags any synchronous `setState` call at the top level of an effect callback. 
  - Already activated, redundant
- **Rule:** When you need to update state from inside an effect, wrap the call in a **named helper function** declared inside the effect instead of calling `setState` inline. 
  - We should discourage use of state on useEffects unless is strictly necessary
- **Rule:** Never call `setState` during the render function body (derived-state-during-render pattern). This can cause render loops when a caller passes a non-stable reference, and React may log warnings.
  - Linter already warms about this
- **Rule:** When you need to derive fresh values from a changed prop **without** waiting for an effect, track the previous prop value in a `useRef` and derive the return value directly during render — mutating a ref does not trigger a re-render.
  - Example incomplete, could encourage pour UX, there is no way of knowing if stats are been requested, or publisher was not provided, also looks like a very specific use case not so valuable.

.github/copilot-instructions.md
```diff
@@ -1671,4 +1679,117 @@ const Component = () => {
...
+
+---
+
+# CI Validation Workflow
+
+These rules define how to validate changes before and after pushing, to avoid wasting CI cycles on issues that can be caught locally.
+
+- **Rule:** Before pushing a branch, always run the linter and tests locally for affected projects. Only push once local checks pass.
+
+```bash
```
johnny-quesada-developer, line 1787: This is misleading... you should run the test for all the projects, we have a mono repo which declare varios projects which then are reuse in others, so green flags at project level doesn't mean that other projects where not affected.

johnny-quesada-developer (changes requested):

OscarFava:
We can delete it by now.

PR #424, opened 2026-03-25
VIDSOL-639: add logger context and update testing document

Hossein-Movahed:
#### What is this PR doing?
Add logger context (`userId`, `sessionId`,`connectionId`)

Adds a `LoggerProvider` that automatically syncs `userId`, `sessionId`, and connectionId into the logger, so we no longer have to pass them manually on every `log()` / `reportError()` call.

What changed:

New `LoggerProvider` sits in the React tree and calls `logger.setContext()` whenever the user or session changes
`LoggerBase` gets `setContext()` / `clearContext()`, context fields are merged into every log call automatically
`userId` added to `ClientLogEvent` schema so it flows through to Kibana
Cleaned up redundant `sessionId`/`connectionId` passing at call sites in `vonageVideoClient.ts` (kept them only in `connect()` due to a React re-render timing gap)
Tests for `LoggerProvider`, `useLoggerContext`, and updated `TESTING.md` with the `makeTestProvider`

Join a meeting room , copy your `sessionId` from the network tab, search for it in Kibaba

in Kibana Search for 
```
sessionId: "YOUR_SESSION_ID"
AND 
action: "EnterMeeting"
```

Resolves [VIDSOL-639](https://jira.vonage.com/browse/VIDSOL-639)

#### Checklist
[ ] Resolves a `Known Issue`.


frontend/src/Context/LoggerProvider/LoggerSynchronizer.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+  },
+}));
+
+const mockSetContext = vi.mocked(frontendLogger.setContext);
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const mockVonageVideoClient = { sessionId: 'session-1', connectionId: 'connection-1' } as any;
+
+describe('useLoggerContext', () => {
+  beforeEach(() => {
```
johnny-quesada-developer: this is not needed

frontend/src/Context/LoggerProvider/LoggerSynchronizer.spec.tsx
```diff
@@ -0,0 +1,125 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
```
johnny-quesada-developer: Hi @Hossein-Movahed ,

Most of the tests are repetitive and low-value. They mainly assert that after mocking the logger, the effect calls the mock with the same values we just fed into the hook. That does not validate meaningful behavior; it mostly validates the mock setup. The suite is growing with tests that do not protect real regressions, which makes it harder to maintain and lowers the overall signal quality while also making the test suit take longer.

The design also does not seem to fit the problem. This goal is to read values from user/session state and synchronizing them into a global logger service. That is a synchronization concern, not a context concern, you don't need context.

This should likely be reduced to a small headless synchronizer, not a provider. One null-rendering component whose only responsibility is to observe values and and sync them into the global logger...

Nothing else. No extra context API, no more renders...

Why a headless synchronizer instead of a hook? To avoid unnecessary re renders on the main tree.

johnny-quesada-developer (changes requested):

frontend/src/Context/LoggerProvider/LoggerSynchronizer.spec.tsx
Hossein-Movahed: @johnny-quesada-developer, changes applied: replaced `LoggerProvider` with headless `LoggerSynchronizer`, removed `LoggerContext`, placed as sibling in `App.tsx`, consolidated tests into one file, removed beforeEach.

OscarFava (dismissed):

frontend/src/logger/transport/backendLogTransport.spec.ts
```diff
@@ -11,6 +12,19 @@ vi.mock('../../utils/getAppVersion', () => ({
   default: () => 'vera-1.0.0-test',
 }));
 
+vi.mock('../event', () => ({
```
johnny-quesada-developer: Don't mock event, mock third parties instead of our own logic.

frontend/src/logger/transport/backendLogTransport.spec.ts
```diff
@@ -67,4 +81,22 @@ describe('BackendLogTransport', () => {
     expect(blob.type).toBe('application/json');
     expect(blob.size).toBeGreaterThan(0);
   });
+
+  it('log() passes userId to createClientEvent when provided in extra', () => {
```
johnny-quesada-developer: these two test do not provide value, you mocked the **/event** and here you are checking that you got back the mocked results... test actual functionality instead

frontend/src/logger/transport/backendLogTransport.ts
```diff
@@ -12,14 +12,15 @@ export class BackendLogTransport {
...
-    const { sessionId, connectionId, partnerId, ...rest } = extra ?? {};
+    const { sessionId, connectionId, partnerId, userId, ...rest } = extra ?? {};
     const clientEvent = {
       ...createClientEvent({
         level: 'info',
         action: event,
         sessionId: typeof sessionId === 'string' ? sessionId : undefined,
         connectionId: typeof connectionId === 'string' ? connectionId : undefined,
         partnerId: typeof partnerId === 'string' ? partnerId : undefined,
+        userId: typeof userId === 'string' ? userId : undefined,
```
johnny-quesada-developer: `userId: typeof userId === 'string' ? userId : undefined,` this guards are totally unnecessary... you still include the key so you can use just { userId }

frontend/src/logger/transport/backendLogTransport.ts
```diff
@@ -46,6 +47,7 @@ export class BackendLogTransport {
...
```
johnny-quesada-developer: [same](https://github.com/Vonage/vonage-video-react-app/pull/424/changes#r3109697469)

libs/common/src/logger/Logger.ts
```diff
@@ -192,6 +203,20 @@ export class LoggerBase implements LoggerProviderConfig {
...
+   */
+  public setContext(context: Partial<LogContext>): void {
+    this.context = { ...this.context, ...context };
+  }
+
+  /**
+   * Clears all persistent context (e.g. on session disconnect or user logout).
+   */
+  public clearContext(): void {
+    this.context = {};
```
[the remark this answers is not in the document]
johnny-quesada-developer: Hi @Hossein-Movahed ,

I think copilot comment is right, the data should be on sync as long as possible and keeping the reference is a simple way for achieving that.

johnny-quesada-developer (changes requested):

frontend/src/Context/LoggerProvider/LoggerSynchronizer.spec.tsx
johnny-quesada-developer: thanks

johnny-quesada-developer (approved):
LGTM

VZaphod (approved):
LGTM

rserebrennykov (approved):
LGTM

PR #452, opened 2026-04-21
VIDSOL-392: remove duplicate "System Default" speaker from speaker dropdown

Hossein-Movahed:
#### What is this PR doing?

Chrome exposes a virtual device with deviceId `"default" `alongside the real output device, causing duplication in the speaker list. Merge the default entry into the matching real device and append `"- System Default"` to its label, matching Google Meet's behaviour. Applied to both the meeting room and waiting room dropdowns via a shared `mergeAudioDeviceLabel` utility.

Browser device enumeration behavior:

### Chrome
Exposes a virtual deviceId: `'default' `entry for `audioinput` and `audiooutput` (not videoinput) with label `"Default "`, alongside the real device, creating a duplicate `mergeAudioDeviceLabel` removes the virtual entry and appends - `System Default` to the matched real device.

### Safari
Exposes a virtual deviceId:

default entry for `audiooutput` only (same "Default - X" format as Chrome), handled by `mergeAudioDeviceLabel`.
Does not expose a default entry for `audioinput` real device IDs and labels only, no suffix needed/expected. 

### Firefox
Returns an empty [], does not expose a default entry for any device kind
 
### Edge
Behaves identically to Chrome, exposes a virtual deviceId: `'default'` entry for both `audioinput` and `audiooutput` (not videoinput), with matching groupId to the current system default physical device. `mergeAudioDeviceLabel` handles both kinds the same way it does for Chrome.

| Browser | `audioinput` default | `audiooutput` default | `videoinput` default |
|--------|----------------------|----------------------|----------------------|
| Chrome | ✅ | ✅ | ❌ |
| Edge | ✅ | ✅ | ❌ |
| Safari | ❌ | ❌ | ❌ |
| Firefox | ❌ | ❌ (empty) | ❌ |
 
In the waiting room or meeting room check the label on speaker dropdown

Resolves [VIDSOL-392](https://jira.vonage.com/browse/VIDSOL-392)

#### Checklist
[ ] Resolves a `Known Issue`.


Hossein-Movahed:
vcr:deploy

Hossein-Movahed:
@johnny-quesada-developer, I don't think moving the label fallback inside `useDistinctLabelMediaDevices` is a good idea, ` device.label ?? t('unknown.device')` . The hook has no translation context, `useTranslation()` only works inside React components bound to the i18n provider. Any fallback the hook provides would have to be either a hardcoded English string or a raw i18n key, both of which break localization.

The current pattern in VideoDevices is correct: the component owns the translation, calls t('unknown.device') to get the locale-appropriate string, and passes it into the selector. The hook stays pure and app-agnostic.

PS. the rest of the suggestion is applied, move the `useDistinctLabelMediaDevices`  to the vera room level, and update the MenuDevices now uses `useDistinctLabelMediaDevices` instead of `useMediaDevices` + manual `cleanAndDedupeDeviceLabels`, matching the pattern already used in `OutputAudioDevices`

johnny-quesada-developer:
> @johnny-quesada-developer, I don't think moving the label fallback inside `useDistinctLabelMediaDevices` is a good idea, ` device.label ?? t('unknown.device')` . The hook has no translation context, `useTranslation()` only works inside React components bound to the i18n provider. Any fallback the hook provides would have to be either a hardcoded English string or a raw i18n key, both of which break localization.
> 
> 
> 
> The current pattern in VideoDevices is correct: the component owns the translation, calls t('unknown.device') to get the locale-appropriate string, and passes it into the selector. The hook stays pure and app-agnostic.
> 
> 
> 
> PS. the rest of the suggestion is applied, move the `useDistinctLabelMediaDevices`  to the vera room level, and update the MenuDevices now uses `useDistinctLabelMediaDevices` instead of `useMediaDevices` + manual `cleanAndDedupeDeviceLabels`, matching the pattern already used in `OutputAudioDevices`

If you moved the useDistinctLabelMediaDevices to the app now you have the i18 context ... there is no reason for repeating code through the app.

Also, as I commented both the core library and the ui library both are going to have language context due a runtime provider we are implementing, but for this specific PR having that logic inside useDistinctLabelMediaDevices is okay.

johnny-quesada-developer:
> > @johnny-quesada-developer, I don't think moving the label fallback inside `useDistinctLabelMediaDevices` is a good idea, ` device.label ?? t('unknown.device')` . The hook has no translation context, `useTranslation()` only works inside React components bound to the i18n provider. Any fallback the hook provides would have to be either a hardcoded English string or a raw i18n key, both of which break localization.
> > The current pattern in VideoDevices is correct: the component owns the translation, calls t('unknown.device') to get the locale-appropriate string, and passes it into the selector. The hook stays pure and app-agnostic.
> > PS. the rest of the suggestion is applied, move the `useDistinctLabelMediaDevices`  to the vera room level, and update the MenuDevices now uses `useDistinctLabelMediaDevices` instead of `useMediaDevices` + manual `cleanAndDedupeDeviceLabels`, matching the pattern already used in `OutputAudioDevices`
> 
> If you moved the useDistinctLabelMediaDevices to the app now you have the i18 context ... there is no reason for repeating code through the app.
> 
> Also, as I commented both the core library and the ui library both are going to have language context due a runtime provider we are implementing, but for this specific PR having that logic inside useDistinctLabelMediaDevices is okay.

Hi @Hossein-Movahed , 

Please try to answer back to the threads once you attended the feedback to avoid misunderstandings.

Then, let's delete the hook **usePreferredCameras** , it doesn't have a reason to exists any longer... it can totally be replace with **useDistinctLabelMediaDevices**

johnny-quesada-developer (changes requested):
delete **usePreferredCameras**

Hossein-Movahed:
> delete **usePreferredCameras**

Good idea, the `useDistinctLabelMediaDevices` covers the same need plus
applies cleaned labels, deduplication, and the unknown-device fallback.
The `usePreferredCameras` is removed

johnny-quesada-developer:
Won't do, The current behavior is expected, chrome handles the concept of default device which can make it look like a duplication but is not. On the other hand Safari and Firefox do not handle the concept of "default device" and emulating the same behaviour will be impossible or too complicated. We are closing this task and the ongoing PR.

The pending changes about adding translations to the default label, they are unrelated to the ticket so will need to reevaluate that later.

PR #459, opened 2026-04-21
VIDSOL-748: Default Screen-Sharing Surface Type (resolves #410)

### LINKED ISSUE #410: [Feature Request]: Choose default media option when screen-scharing
### Is this a known issue?

- [x] I have searched the existing issues

### Describe the feature request

Hi there,

I would like to submit a feature request regarding the screen-sharing behavior when using **OT.initPublisher**.

Currently, when a user starts screen sharing, the browser prompts them to choose which media to share. By default, the selection opens on the **browser tab** option, while the user can also switch to **window** or **entire screen**.

Using the standard Web API, it is possible to control which option is pre-selected in the browser picker through the **displaySurface** constraint in **MediaTrackConstraints**. For example:

```
// Pre-select the "Entire screen" pane in the media picker
const media = await navigator.mediaDevices.getDisplayMedia({
  video: { displaySurface: "monitor" }
});

```
Reference:
https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints/displaySurface#value

However, this configuration does not appear to be accessible when using **OT.initPublisher**.

In my use case, all my users are choosing **entire screen** when screen-sharing, so being able to configure the default selection would significantly improve the user experience during screen sharing.

Would it be possible to expose this option (for example via MediaTrackConstraints or a similar parameter) within the initPublisher method?

Thank you for considering this request.

Best regards,
Arnaud

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

czoli1976:
#### What is this PR doing?

Adds a new optional frontend env var `DEFAULT_SCREEN_SHARE_SURFACE` that pre-selects which tab of the browser's screen-share picker is shown when a user starts sharing (`monitor` → *Entire screen*, `window` → *Window*, `browser` → *Chrome Tab*). Unset keeps today's behaviour — the browser default.

**How it works.** The Vonage SDK's `initPublisher({ videoSource: 'screen' })` path does not expose the [`displaySurface`](https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints/displaySurface) constraint that the Web API supports. The SDK does, however, accept a `constraints` property on `PublisherProperties` and forwards `constraints.video` verbatim to `getDisplayMedia`. We attach the user's chosen `displaySurface` through that path while keeping `videoSource: 'screen'`, so the stream is still classified as `videoType === 'screen'` — remote subscribers, the `screenshareStreamCreated` event, and all existing `videoType === 'screen'` checks are unaffected.

The setting is deployment-wide (no per-user UI) — the operator picks the best default for their audience in `vcrBuild.env.sh`.

Works best in Chromium (Firefox/Safari honour the hint partially or not at all, per [spec](https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints/displaySurface#browser_compatibility) — that's acceptable).

| Step | Expected |
|---|---|
| Leave `DEFAULT_SCREEN_SHARE_SURFACE` unset in `vcrBuild.env.sh`, `yarn dev`, join a room, click **Share screen** | Picker opens on the browser's default tab (usually *Chrome Tab*). Unchanged behaviour. |
| Set `export DEFAULT_SCREEN_SHARE_SURFACE='monitor'`, restart, click **Share screen** | Picker opens on *Entire Screen*. |
| Repeat with `'window'` | Picker opens on *Window*. |
| Repeat with `'browser'` | Picker opens on *Chrome Tab*. |
| Set `='bogus'`, restart | App startup throws a validation error — invalid values are rejected at load time. |
| From a second browser, join the same room while the first is sharing | Shared stream appears as a screen-share tile (not a camera tile). Confirms `videoType === 'screen'` is preserved. |

Resolves [VIDSOL-748](https://jira.vonage.com/browse/VIDSOL-748) and [Vonage/vonage-video-react-app#410](https://github.com/Vonage/vonage-video-react-app/issues/410).

#### Checklist

- [x] Branch is based on `develop` (not `main`)
- [ ] Resolves a `Known Issue`
- [ ] If yes, did you remove the item from the `docs/KNOWN_ISSUES.md`?
- [x] Resolves an item reported in `Issues`

Generated with [Claude Code](https://claude.com/claude-code)


frontend/src/hooks/tests/useScreenShare.spec.tsx
```diff
@@ -264,6 +265,69 @@ describe('useScreenSharing', () => {
...
+      });
+
+      expect(initPublisher).toHaveBeenCalledWith(
+        undefined,
+        expect.objectContaining({
+          videoSource: 'screen',
+          constraints: { video: { displaySurface: surface }, audio: false },
+        }),
+        expect.any(Function)
+      );
```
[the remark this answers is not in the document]
czoli1976: Leaving this one as-is — env is already reset between tests by the global `afterEach(() => env.reset())` in [`frontend/src/test/setup.ts:11-13`](https://github.com/Vonage/vonage-video-react-app/blob/feature/screen-share-display-surface/frontend/src/test/setup.ts#L11-L13), so the singleton can't leak into the "env is unset" case regardless of test order. Matches the pattern used by the existing `env.partialUpdate` callers elsewhere in the suite (e.g. [`WaitingRoom.spec.tsx`](https://github.com/Vonage/vonage-video-react-app/blob/develop/frontend/src/pages/WaitingRoom/WaitingRoom.spec.tsx), [`ScreenSharingButton.spec.tsx`](https://github.com/Vonage/vonage-video-react-app/blob/develop/frontend/src/components/ScreenSharingButton/ScreenSharingButton.spec.tsx)), none of which use `try/finally`.

frontend/src/hooks/useScreenShare.tsx
```diff
@@ -1,8 +1,9 @@
 import { useState, useRef, useCallback } from 'react';
-import { Publisher, initPublisher } from '@vonage/client-sdk-video';
+import { Publisher, PublisherProperties, initPublisher } from '@vonage/client-sdk-video';
```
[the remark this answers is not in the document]
czoli1976: Addressed in [b03e799](https://github.com/Vonage/vonage-video-react-app/pull/459/commits/b03e7993e228ea185a95f511cd58814dccef7b1d) — split into `import { initPublisher }` + `import type { Publisher, PublisherProperties }`.

frontend/src/hooks/tests/useScreenShare.spec.tsx
```diff
@@ -6,6 +6,7 @@ import { makeTestProvider, providers, ProviderOptions } from '@test/providers';
 import EventEmitter from 'events';
 import type VonageVideoClient from '../../utils/VonageVideoClient';
 import { type UserContextType } from '../../Context/user';
+import { env } from '../../env';
```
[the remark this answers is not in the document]
czoli1976: Addressed in [b03e799](https://github.com/Vonage/vonage-video-react-app/pull/459/commits/b03e7993e228ea185a95f511cd58814dccef7b1d) — `Publisher`, `ProviderOptions`, `UserContextType`, and `VonageVideoClient` are now imported with `import type`. `initPublisher` stays as a value import so the `vi.mock` stub keeps working.

frontend/src/types/session.ts
```diff
@@ -58,3 +58,7 @@ export type StreamPropertyChangedEvent = {
 export const LAYOUT_MODES = ['grid', 'active-speaker'] as const;
 
 export type LayoutMode = (typeof LAYOUT_MODES)[number];
+
+export const DISPLAY_SURFACES = ['monitor', 'window', 'browser'] as const;
+
```
Hossein-Movahed: `DISPLAY_SURFACES` and `DisplaySurface` doesnt belong to session, they should be part of screenShare domain

frontend/src/hooks/useScreenShare.tsx
```diff
@@ -17,6 +19,12 @@ export type UseScreenShareType = {
   screenshareVideoElement: HTMLVideoElement | HTMLObjectElement | undefined;
 };
 
+// `constraints` isn't declared on the SDK's PublisherProperties, but the SDK
+// forwards it verbatim to getDisplayMedia — letting us hint the picker tab.
+type ScreenSharePublisherProperties = PublisherProperties & {
```
Hossein-Movahed: [Same](https://github.com/Vonage/vonage-video-react-app/pull/459/changes#r3132108671)

frontend/src/env.ts
```diff
@@ -107,6 +112,22 @@ function parseResolution(value: unknown, fallback: Resolution | undefined): Reso
   return value as Resolution;
 }
 
+function parseDisplaySurface(value: unknown): DisplaySurface | undefined {
```
Hossein-Movahed: `value: unknown` is too generic, how about `rawValue`?

frontend/src/hooks/useScreenShare.tsx
```diff
@@ -59,21 +67,26 @@ const useScreenShare = (): UseScreenShareType => {
...
+          videoSource: 'screen',
+          insertDefaultUI: false,
+          videoContentHint: 'detail',
+          name: t('participants.screen', { participantName: user.defaultSettings.name }),
+        };
+
+        if (env.DEFAULT_SCREEN_SHARE_SURFACE) {
+          publisherProperties.constraints = {
+            video: { displaySurface: env.DEFAULT_SCREEN_SHARE_SURFACE },
+            audio: false,
```
Hossein-Movahed: ` audio: false`  <- who decided this? 
This silently disables screen share audio when `DEFAULT_SCREEN_SHARE_SURFACE ` is set. Operator sets the env var to control the picker tab, not to disable audio. These are two orthogonal concerns

frontend/src/hooks/useScreenShare.tsx
```diff
@@ -59,21 +67,26 @@ const useScreenShare = (): UseScreenShareType => {
...
       if (!isSharingScreen) {
+        const publisherProperties: ScreenSharePublisherProperties = {
+          videoSource: 'screen',
+          insertDefaultUI: false,
+          videoContentHint: 'detail',
+          name: t('participants.screen', { participantName: user.defaultSettings.name }),
+        };
+
+        if (env.DEFAULT_SCREEN_SHARE_SURFACE) {
+          publisherProperties.constraints = {
```
Hossein-Movahed:
```
if (env.DEFAULT_SCREEN_SHARE_SURFACE) {
  publisherProperties.constraints = { ... };  // mutating after declaration
}
```

Not wrong, but the reader has to scan past the declaration to know what the final object looks like. A spread is cleaner and removes the optional-property type entirely.

`ScreenSharePublisherProperties` type can be dropped entirely, TypeScript infers the correct shape.
`PublisherProperties` is imported but only used in the `ScreenSharePublisherProperties` intersection. If you drop that type, `PublisherProperties` import goes away too , one less thing to maintain.

frontend/src/hooks/useScreenShare.tsx
```diff
@@ -1,8 +1,10 @@
 import { useState, useRef, useCallback } from 'react';
-import { Publisher, initPublisher } from '@vonage/client-sdk-video';
+import { initPublisher } from '@vonage/client-sdk-video';
+import type { Publisher, PublisherProperties } from '@vonage/client-sdk-video';
```
Hossein-Movahed: Could be :
`import { initPublisher, type Publisher, type PublisherProperties } from '@vonage/client-sdk-video';`

Hossein-Movahed (changes requested):

frontend/src/types/session.ts
czoli1976: Done in [2ff6843](https://github.com/Vonage/vonage-video-react-app/pull/459/commits/2ff684351fa6e5b46e8df0e3554b3d4afe3712db) — moved `DISPLAY_SURFACES` and `DisplaySurface` to a new `frontend/src/types/screenShare.ts` module and updated the `env.ts` import accordingly.

frontend/src/hooks/useScreenShare.tsx
czoli1976: Done in [2ff6843](https://github.com/Vonage/vonage-video-react-app/pull/459/commits/2ff684351fa6e5b46e8df0e3554b3d4afe3712db) — same move as the session.ts comment; the type now lives in `frontend/src/types/screenShare.ts`.

frontend/src/env.ts
czoli1976: Done in [2ff6843](https://github.com/Vonage/vonage-video-react-app/pull/459/commits/2ff684351fa6e5b46e8df0e3554b3d4afe3712db) — renamed to `rawValue`.

frontend/src/hooks/useScreenShare.tsx
czoli1976: Good catch — done in [2ff6843](https://github.com/Vonage/vonage-video-react-app/pull/459/commits/2ff684351fa6e5b46e8df0e3554b3d4afe3712db). Dropped `audio: false` so the constraints now only carry `video.displaySurface`. The SDK forwards `constraints.audio` to `getDisplayMedia` verbatim, so leaving it out matches the prior `videoSource: 'screen'`-only behaviour rather than overriding it.

frontend/src/hooks/useScreenShare.tsx
czoli1976: Done in [2ff6843](https://github.com/Vonage/vonage-video-react-app/pull/459/commits/2ff684351fa6e5b46e8df0e3554b3d4afe3712db) and finalised in [3554046](https://github.com/Vonage/vonage-video-react-app/pull/459/commits/35540463080677ad85eb6ab5d258ce0eedcc6a54).

The first commit kept the literal as a separate `const` and used `as const` to recover the literal types after dropping the alias — but SonarCloud flagged the assertions (`typescript:S4325`).

The second commit inlines the literal at the `initPublisher` call site instead. That gets contextual typing from the SDK signature for free, so no `as const`, no `PublisherProperties` import, no `ScreenSharePublisherProperties` alias. The optional `constraints` field rides in via spread so the excess-property check doesn't fire on a field the SDK accepts but doesn't declare.

`PublisherProperties` import is now gone from this file.

frontend/src/hooks/useScreenShare.tsx
czoli1976: Done in [2ff6843](https://github.com/Vonage/vonage-video-react-app/pull/459/commits/2ff684351fa6e5b46e8df0e3554b3d4afe3712db). After the call-site inline refactor in [3554046](https://github.com/Vonage/vonage-video-react-app/pull/459/commits/35540463080677ad85eb6ab5d258ce0eedcc6a54), `PublisherProperties` is no longer needed at all — the import is now just `import { initPublisher, type Publisher } from '@vonage/client-sdk-video';`.

Hossein-Movahed:
Hey @czoli1976 , develop had some update since last week, please rebase before we continue.

czoli1976:
@Hossein-Movahed Done — rebased onto latest `upstream/develop` (`5edb19c1`); the 5 commits replayed cleanly with no conflicts and ts-check / unit tests / lint all stay green locally. Branch head is now `566f7ba5`.

Hossein-Movahed (dismissed):
Awesome work 🔥

johnny-quesada-developer:
vcr:deploy

frontend/src/env.ts
```diff
@@ -107,6 +108,22 @@ function parseResolution(value: unknown, fallback: Resolution | undefined): Reso
   return value as Resolution;
 }
 
+function parseDisplaySurface(rawValue: unknown): DisplaySurface | undefined {
+  if (rawValue === undefined || rawValue === null || rawValue === '') return undefined;
```
johnny-quesada-developer:
```tsx
if(isNil(rawValue)) return
```

johnny-quesada-developer:
Hi @czoli1976 ,

Thanks for the great job, questions:

A - The actual problem is that the @vonage/client-sdk-video types are out of sync with the internals of **opentok.js.** The solution here is updating @vonage/client-sdk-video so that clients can see that the option is available. This is the critical actual fix.

B - Then, is the client using a fork of Vera? Becase if so, I understand that adding the example on the Vera app could help them, but if they have their own custom code the [PR: VIDSOL-748: Default Screen-Sharing Surface Type (resolves #410)](https://github.com/Vonage/vonage-video-react-app/pull/459/changes#diff-cd0db15b235f8ccec86d550d3c929afbd0fec9f198b7f094bf34ca419d1379c9) won't give them any value.

Could you help me please answering to these points?

cc: @Hossein-Movahed, @VZaphod

johnny-quesada-developer (changes requested):

johnny-quesada-developer:
Hi @czoli1976 , 

This PR is too large and introduces a very significant dependency footprint into the project.

The raise hand feature itself is needed, and we definitely want to have it. However, it should be submitted in an isolated PR so we can review the feature properly on its own, without mixing it with the TensorFlow / AI-based gesture recognition implementation.

Adding TensorFlow to the project is not a small technical detail. It has major architectural and performance implications, and may not be appropriate for VERA, at least not now.

TensorFlow is a powerful library, but it is also a very heavy dependency. Introducing it for this use case would significantly increase the bundle size, memory usage, parsing time, startup time, and runtime cost of the application. Even while testing we spot a couple of performance issues and crashes even on desktop, so adding this kind of client-side ML runtime could degrade the user experience especially for mobile users, or bad connections.

There is also a distribution problem here. The current implementation adds dependencies directly from a CDN, which is problematic:

- it bypasses our normal build pipeline
- it prevents proper dependency deduplication and optimization
- it makes versioning and reproducibility harder
- it creates an additional reliability risk
- it increases the security and supply-chain attack surface
- it may fail in client environments where external resources are blocked

This is especially concerning for the prebuilt and embeddable versions of the application. For those builds, we are trying to provide self-contained, lightweight bundles that clients can include in their own websites in a single one file bundle.

A 40-50 MB bundle may not sound huge in isolation, but it is far too big for a modern web app or embeddable web component. Regular websites bundles are about to 10-13 MB, the current Vera bundle is 4 MB... adding this feature with TensorFlow will push us to around 40+ MB for 1 feature, which doesn't sounds like the correct path. 

For embeddable versions, this is even more critical because we do not control the client environment. Some clients may block CDN downloads, restrict dynamic module loading, have strict CSP rules, or run the component in constrained devices or networks. We should not introduce a dependency model that makes integration harder or less predictable.

From a product perspective, this also feels misaligned. Major video conferencing products such as Zoom, Google Meet, and similar providers do not rely on shipping a large client-side TensorFlow runtime just to support basic meeting interaction features. There is a reason for that: the performance and distribution cost outweighs the value for this kind of feature.

The raise hand functionality should remain a normal explicit user action in the UI. It should not require a heavy AI/ML runtime.

**Requested change**
Please split this PR and keep the raise hand feature isolated from the TensorFlow / AI gesture recognition implementation.

- The raise hand feature is needed, but the TensorFlow-based detection is not aligned with current Vera goals. It adds a very large dependency footprint, increases performance risk, and conflicts with our current delivery.
- This new PR should not introduce TensorFlow, CDN-loaded dependencies, or client-side ML for gesture recognition. 
- The scope should be limited to the core raise hand feature only.

cc: @Hossein-Movahed , @VZaphod

frontend/src/env.ts
czoli1976: Done in [9097dae](https://github.com/Vonage/vonage-video-react-app/pull/459/commits/9097daecf2b7ac024e8d8e2d9fca7e9e4d26bfe5). Using `isNil` from `@common/assertions/isNil`. Kept the `rawValue === ''` check alongside it since `isNil('')` returns `false` — empty string should also be treated as unset, consistent with the pattern in `parseResolution`.

czoli1976:
Hi @johnny-quesada-developer, thanks for the questions!

**A** — Agreed, updating `@vonage/client-sdk-video` types to declare `constraints` on `PublisherProperties` is the right long-term fix and would let us drop the spread workaround. That's squarely in the SDK team's lane though — happy to file a separate issue if that helps. This PR is intentionally scoped to the reference app; the code comment documents the workaround explicitly so it's easy to remove once the types land.

**B** — Vera serves two audiences: customers who deploy it directly (this env var lets them configure the picker tab without touching code) and developers who use it as a reference template (they can copy the pattern). Issue #410 was opened by a customer wanting exactly this behaviour in the deployed app, which is why we targeted it here.

czoli1976:
@johnny-quesada-developer thanks for the careful read — fair concerns. A couple of corrections on the framing, then I'm fully on board with the split.

**On the dependency**

It's **MediaPipe Tasks** (TF Lite XNNPACK runtime), not full TensorFlow.js — much smaller surface than the comment implies.

More importantly: **the Vonage Video Client SDK already bundles `@mediapipe/tasks-vision`** and pulls it from the same CDN. Grepping `node_modules/@vonage/client-sdk-video/dist/js/opentok.js`:

```
@mediapipe/tasks-vision
cdn.jsdelivr.net/npm/@mediapipe/tasks-vision
mediapipe/modules/face_detection/
mediapipe/modules/hand_landmark/        ← same model family our recogniser uses
mediapipe/modules/pose_landmark/
mediapipe/modules/selfie_segmentation/
ImageSegmenter / imageSegmenter / mediapipeBaseAssetsUri
```

Background blur and replacement (`hasMediaProcessorSupport()` → `setVideoFilter`) already use this runtime. So every VERA user with background effects enabled — which is on by default — is already downloading MediaPipe WASM + a model from the SDK. The gesture-detection PR is adding *a second copy* of the same runtime, which is the actual waste.

**On the size**

The "40–50 MB" figure is from before the self-host commit. After commit `58e0481d` the **wire transfer** numbers are:

| File | Raw | Brotli q11 |
|---|---|---|
| `vision_wasm_internal.wasm` | 11.0 MB | 2.2 MB |
| `vision_wasm_nosimd_internal.wasm` | 10.0 MB | 2.2 MB |
| `gesture_recognizer.task` (float16) | 8.0 MB | 5.6 MB |
| 2× `.js` glue | 400 KB | 84 KB |
| **First-load total** | **29.4 MB** | **10.1 MB** |

And it's gated behind a dynamic `import()` so users who never enable gesture detection ship 0 KB of MediaPipe app-side. The CDN concerns (no SRI, offline, supply chain) are fixed in the same commit — assets are self-hosted from `frontend/public/mediapipe/`, pre-compressed at build time, served with `Content-Encoding: br` via `express-static-gzip`.

**The right architectural fix**

Given the SDK already loads MediaPipe for blur/replacement, the cleanest design is to expose gesture recognition as a first-class media-processor capability on `client-sdk-video` — mirror of `setVideoFilter`:

```ts
publisher.setGestureRecognizer({
  gestures: ['Open_Palm', 'Thumb_Up', 'Thumb_Down'],
  intervalMs: 250,
  confidence: 0.45,
  onGesture: (gesture) => { /* … */ },
});
```

Reuses the SDK's already-loaded WASM runtime, only the second `.task` model file would be added. VERA app code shrinks to ~30 lines: subscribe and call `raiseHand` / `sendEmoji`. No `@mediapipe/tasks-vision` direct dep, no CDN handling, no self-hosting script in the app repo.

**Path forward — agree with the split**

1. Carve PR #462 into PR A (raise-hand only) and PR B (gesture detection). Ship A.
2. File a `client-sdk-video` proposal for `setGestureRecognizer` so the SDK's MediaPipe instance is shared.
3. Revisit gesture detection in VERA after the SDK API lands — VERA becomes a thin consumer instead of bundling a second ML runtime.

Working on the split now. cc @VZaphod — interested in your take on whether the SDK is the right home for gesture recognition.

czoli1976:
@johnny-quesada-developer just to close the loop on this — we've done the split:

- **PR #462** is now **raise-hand only** (commit `93bf709c`, renamed `VIDSOL-645: Raise Hand`). All MediaPipe / gesture detection / CDN concerns removed: no `@mediapipe/tasks-vision` dep, no self-hosted assets, no `express-static-gzip`, no gesture-ring CSS, no postinstall network call. 1020 frontend tests passing. [Status summary on #462](https://github.com/Vonage/vonage-video-react-app/pull/462#issuecomment-4386448196).
- The full gesture-detection work is preserved on a separate upstream branch [`VIDSOL-645-ai-gestures`](https://github.com/Vonage/vonage-video-react-app/tree/VIDSOL-645-ai-gestures) so nothing is lost. We'll only open a follow-up PR after the SDK conversation about `publisher.setGestureRecognizer({...})` (sharing the SDK's already-loaded MediaPipe instance instead of bundling a second one).

Appreciate the pushback — the split is genuinely the right move and the SDK-vs-app architectural question is worth getting right before re-introducing the feature.

johnny-quesada-developer (approved):

VZaphod:
This is a type error in the Video Client SDK init publisher parameters declaration, we don't see the value in extending VERA to add this feature right now. We'll consider doing something similar in the incoming user preferences view. It makes more sense to have it there. A preference to select the default screen sharing surface.

czoli1976:
Expected a fix in the SDK that should make this  kind-of-an-hack PR not neeed anymore.

PR #460, opened 2026-04-22
VIDSOL-727: Remove mui theme, added parameters to generate vera UI plugin, fixes on vera web client now uses context, added vera studio local

johnny-quesada-developer:
#### What is this PR doing?
image
https://neru-412f5c1c-vonage-video-react-app-vera-dev.euw1.runtime.vonage.cloud/

Resolves [VIDSOL-727](https://jira.vonage.com/browse/VIDSOL-727)

#### Checklist
[ ] Resolves a `Known Issue`.


johnny-quesada-developer:
vcr:deploy

VZaphod:
Join room is broken in https://neru-412f5c1c-vonage-video-react-app-vera-dev.euw1.runtime.vonage.cloud

frontend/src/components/MeetingRoom/Emoji/Emoji.tsx
```diff
@@ -40,25 +38,20 @@ const Emoji = ({ emojiWrapper }: EmojiProps): ReactElement => {
         ml: isSmallViewport ? 5 : '15%',
         display: 'flex',
         flexDirection: 'column',
-        fontSize: isSmallViewport
-          ? theme.typography.typeScale.mobile['subtitle'].fontSize.value
-          : theme.typography.typeScale.desktop['subtitle'].fontSize.value,
+        fontSize: isSmallViewport ? '1.875rem' : '3.25rem',
```
VZaphod: Magic numbers? shouldn't be constants? should we have a viewportBased fontSize computed property?

backend/middleware/errorHandler.ts
```diff
@@ -65,9 +65,11 @@ export function errorHandler(
   const isHtmlRequest = accepts.includes('text/html');
 
   if (isHtmlRequest) {
-    res.status(statusCode).render('index', {
-      error: safeError,
-    });
+    res
+      .status(statusCode)
+      .send(
+        `<!DOCTYPE html><html><body><h1>${statusCode}</h1><p>${safeError.message}</p></body></html>`
```
Hossein-Movahed: This is a reflected XSS vulnerability. `safeError.message`  interpolated raw into HTML. Old `render()` escaped it automatically; new `.send()` doesn't. Need `escapeHtml()` wrapper. This could be a security issue.
Wrap: `safeError.message.replace(/&/g,'&amp;').replace(/</g,'&lt;').replace(/>/g,'&gt;')` or pull in he / escape-html.

frontend/src/VeraRoom/hooks/useStateSynchronizer.ts
```diff
@@ -0,0 +1,52 @@
...
+          skipFirst: true,
+        }
+      ),
+
+      // clientUrl changes from the bridge should update the video client in the runtime store
+      bridge.subscribe(
+        ({ entryPoint }) => entryPoint,
+        (entryPoint) => {
+          runtime.setState((state) => ({
+            ...state,
```
Hossein-Movahed, line 34: The first value of `entryPoint` is skipped because it's assumed to be already set via `runtime$.Provider` in `VeraRoomElement`. But if this hook is reused in a different context (without that Provider pre-seeding the value), the initial `entryPoint` will never trigger a `videoClient` creation. This coupling is implicit and fragile.

libs/core/src/stores/runtime/RuntimeProvider.ts
```diff
@@ -0,0 +1,63 @@
...
+import { Prettify } from '@common/types';
+import { useStableRef } from '@web/hooks';
+import { QueryClient } from '@tanstack/react-query';
+import { type VideoClient, createVideoClient } from '@core/services';
+import { isString } from '@common/assertions';
+
+type RuntimeState = {
+  language?: string;
+  queryClient?: QueryClient;
+  videoClient: VideoClient | string;
```
Hossein-Movahed, line 12: `VideoClient` and string are semantically unrelated under same prop. Consider `videoClientUrl?: string` + `videoClient?: VideoClient` with mutual exclusion enforced by type.

frontend/src/VeraRoom/stages/MeetingRoomStage/MeetingRoomStage.tsx
```diff
@@ -9,7 +9,7 @@ import MeetingRoom from '@pages/MeetingRoom';
  * Embeddable version of the meeting room. Provides SessionProvider and PublisherProvider.
  */
 const MeetingRoomStage: FC = () => (
-  <SessionProvider>
+  <SessionProvider skipBrowserUrlUpdate>
```
Hossein-Movahed, line 12: Hardcoded true, if `MeetingRoomStage` ever renders outside embedded context, URL won't update and nobody will know why. Drive from context or parent prop rather than hardcoding.

frontend/src/VeraRoom/styles.css
```diff
@@ -1,5 +1,13 @@
 @import 'tailwindcss';
 
+/* Keeps output pixel exact after migration from muiV5 to muiV9 */
+@layer base {
+  * {
+    letter-spacing: 0.15008px;
```
Hossein-Movahed, line 6: NIT: Global * with magic pixel value will silently override any component that sets its own `letter-spacing` (inputs, headings, code). Scope to `.vera-room-root` to contain the blast radius.

frontend/src/VeraRoom/VeraRoomElement.tsx
```diff
@@ -81,9 +88,11 @@ class VeraRoomElement extends HTMLElement {
 
     this.root?.render(
       <BridgeProvider value={initialState}>
-        <ShadowStylesProvider shadowRoot={this.shadow}>
-          <VeraRoom />
-        </ShadowStylesProvider>
+        <runtime$.Provider videoClient={initialState.entryPoint} language={initialState.language}>
+          <ShadowStylesProvider shadowRoot={this.shadow}>
```
Hossein-Movahed, line 92: `initialState.entryPoint` is a string (URL), but `videoClient` prop expects a `VideoClient` object. This will silently fail at runtime. The web component `attributeChangedCallback` later syncs via `useStateSynchronizer`, but the initial value is wrong.

Possible fix: Use `createVideoClient({ url: initialState.entryPoint })` here, or pass null/undefined and let the synchronizer set it.

frontend/src/Context/SessionProvider/session.tsx
```diff
@@ -488,14 +496,16 @@ const SessionProvider = ({ children, initialValue = {} }: SessionProviderProps):
...
+        window.history.replaceState(null, '', `/room/${args.sessionKey}`);
+      }
 
       return connect({
         sessionKey: args.sessionKey,
         token: session.token,
       });
     },
-    [connect]
+    [connect, videoClient, skipBrowserUrlUpdate]
```
Hossein-Movahed, line 508: `videoClient` was added to the dependency array, but from the visible diff, it's not actually used inside the callback body. This will cause `joinSession` to be recreated every time the video client context changes, potentially disrupting in flight session logic.

frontend/src/components/LanguageSelector/LanguageSelector.tsx
```diff
@@ -20,13 +19,15 @@ const languageOptions: LanguageOption[] = [
...
-  <VividIcon name="chevron-down-line" customSize={-5} sx={{ color }} {...props} />
+const ChevronIcon = ({ className, ...props }: { className: string } & Record<string, unknown>) => (
+  <VividIcon name="chevron-down-line" customSize={-5} className={className} {...props} />
 );
 
 const SelectIconComponent =
-  (themeColor: string) =>
-  (props: Record<string, unknown>): ReactElement => <ChevronIcon color={themeColor} {...props} />;
+  (iconClassName: string) =>
+  (props: Record<string, unknown>): ReactElement => (
```
Hossein-Movahed, line 28: `Record<string, unknown>` is an escape hatch. Prefer:
```
React.ComponentPropsWithoutRef<typeof VividIcon>
```

frontend/src/VeraRoom/VeraRoom.spec.tsx
```diff
@@ -1,11 +1,19 @@
...
+import { runtime$ } from '@core/stores';
 import VeraRoom from './VeraRoom';
 import { setupWindowNavigatorMock } from '@web-test/fixtures';
+import { Any } from '@common/types';
 
 const renderWithBridge = (ui: React.ReactElement) =>
-  render(<bridge$.Provider>{ui}</bridge$.Provider>);
+  render(
+    <bridge$.Provider>
+      <runtime$.Provider videoClient={{} as Any}>{ui}</runtime$.Provider>
```
Hossein-Movahed: Even in tests, as `Any` bypasses the compiler. Use the proper mock shape or a typed partial:
```
import type { VideoClient } from '@core/services';
const mockVideoClient = {} as VideoClient;
```

frontend/src/VeraRoom/VeraRoom.stories.tsx
```diff
@@ -4,6 +4,9 @@ import { ComponentProps, memo, PropsWithChildren, useEffect, useRef, forwardRef
 import '../i18n';
 import './VeraRoomElement';
 import { BridgeAttribute } from './stores/bridge';
+import { env } from '../env';
+
+const entryPoint = `${env.API_URL}/v2`;
```
Hossein-Movahed, line 9: This executes at import time. If `env.API_URL` is `undefined` in CI/test environments, stories will silently render with a broken URL. Compute it lazily inside the story render function or use a safe default.

Hossein-Movahed (changes requested):
Thanks for the awesome work Johnny 🦾 , few comments

frontend/src/components/MeetingRoom/Emoji/Emoji.tsx
johnny-quesada-developer: Hi @ivan-vonage ,

Good catch, this one actually we have it as semantic token.

```tsx
 className="text-vera-subtitle"
```

image

For the other values I cleaned a little as well migrating what was possible to migrate to tailwind... It looks like this was the only component that was hard coding this values.

frontend/src/VeraRoom/hooks/useStateSynchronizer.ts
johnny-quesada-developer: This assumption is wrong, the the video client is created on the root runtime provider. This synchronizer is only for the embed versions because the entry point could be updated from the html attributes

libs/core/src/stores/runtime/RuntimeProvider.ts
johnny-quesada-developer: Fair point, but in this case the video client could also be injected from html attributes by using only string... in plain html you cannot pass down a object as parameter. This is basically a method overload, and the result from using string or client is identical.

frontend/src/VeraRoom/stages/MeetingRoomStage/MeetingRoomStage.tsx
johnny-quesada-developer: MeetingRoomStage should never be use outside embedded context... But just in case, this is also a temporal parameter, right now the majority of Vera code is also Embed Room code, so this is one of the tradeoffs we have now, we should be able to tell apart the environments... as we advance with the isolation of the primitives into packages, Vera and Vera Room codes will be separated giving us more freedom to and control.

frontend/src/VeraRoom/styles.css
johnny-quesada-developer: As the comment says, this is hard code to ensure pixel exact migration after updating **mui**... **muiV5** was adding this values and **muiV9** doesn't. Removing it will affect the current design. 

In general components should not randomly add their own **letter-spacing**, we are not doing it because that will break the harmony of the design system.

frontend/src/VeraRoom/VeraRoomElement.tsx
johnny-quesada-developer: The parameter accept both the url or a client instance.

frontend/src/Context/SessionProvider/session.tsx
johnny-quesada-developer: image
Is in use

frontend/src/VeraRoom/VeraRoom.spec.tsx
johnny-quesada-developer: This is just a render test, we have plenty of other test which assert client functionality, but Ill change the parameter.

frontend/src/VeraRoom/VeraRoom.stories.tsx
johnny-quesada-developer: wrong assumption... locally when running stories env will be full filled and working, and CIs they do not run stories.

frontend/src/components/MeetingRoom/ToolbarButton/ToolbarButton.tsx
```diff
@@ -23,11 +23,10 @@ const ToolbarButton = forwardRef(function ToolbarButton(
   ref: ForwardedRef<HTMLButtonElement>
 ) {
   const { className, icon: Icon, sx = {}, isOverflowButton, ...rest } = props;
-  const theme = useTheme();
 
   return (
     <IconButton
-      className={className}
+      className={classNames('bg-vera-dark-grey! hover:bg-vera-dark-grey-opacity', className)}
       {...rest}
```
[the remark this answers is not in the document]
johnny-quesada-developer: copilot comment is wrong, this is valid syntax on the new tailwind.

frontend/src/components/LanguageSelector/LanguageSelector.tsx
```diff
@@ -20,13 +19,15 @@ const languageOptions: LanguageOption[] = [
...
+    <ChevronIcon className={iconClassName} {...props} />
+  );
```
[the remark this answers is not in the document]
johnny-quesada-developer: we remove those parameters they were not necessary.

frontend/src/components/MeetingRoom/Emoji/Emoji.tsx
johnny-quesada-developer: Done.

backend/middleware/errorHandler.ts
johnny-quesada-developer: Done.

frontend/src/VeraRoom/VeraRoom.spec.tsx
johnny-quesada-developer: Done.

Hossein-Movahed (dismissed):
LGTM 🔥

Hossein-Movahed:
vcr:deploy

VZaphod (dismissed):
LGTM

VZaphod (dismissed):
LGTM

Hossein-Movahed (dismissed):

Hossein-Movahed (approved):

johnny-quesada-developer:
vcr:deploy

VZaphod (approved):
LGTM

rserebrennykov (approved):
LGTM

PR #507, opened 2026-05-27
VIDSOL-441: Add-non-available-camera-notification

johnny-quesada-developer:

Resolves [VIDSOL-441](https://jira.vonage.com/browse/VIDSOL-441)

#### Checklist
[ ] Resolves a `Known Issue`.


johnny-quesada-developer:
## PR Overview

This PR refactors the device selection flow so that the actual behavior is handled at the lower-level `selectDevice` path, while components are simplified to call into that shared mechanism. It also improves user feedback for unavailable devices through translated error messaging, removes unused media source settings, and eliminates duplicated logic that had been spread across contexts and UI components.

### Main changes

#### Device selection is centralized at the `selectDevice` level
A key improvement in this PR is that device-switching behavior is now driven from the lower-level `selectDevice` flow rather than being partially implemented across multiple UI components and contexts.

This improves the structure of the codebase by:
- establishing a single path for device selection behavior
- reducing the amount of logic owned by components
- making the behavior more consistent across the app
- making future changes easier to implement in one place

The `useSelectDeviceHandler` hook supports this flow, but its role is mainly to add translated feedback and error presentation. The underlying selection behavior is handled by the shared lower-level path.

#### Components are simplified
Several components are now leaner because they no longer need their own custom device-selection handlers and can instead rely on the shared selection flow.

This simplification affects areas such as:
- `MenuDevices`
- `InputAudioDevices`
- `OutputAudioDevices`
- `VideoDevices`
- waiting room controls
- mobile camera switching from `SmallViewportHeader`

The result is:
- less prop passing
- less duplicated event-handling logic
- clearer separation between UI concerns and device-selection behavior
- easier-to-follow component responsibilities

#### Unused user settings for media sources were removed
The PR also removes unused persisted/default user settings related to device sources, including:
- `audioSource`
- `videoSource`

It also removes related context-level APIs and state that were no longer necessary.

This cleanup makes the state model clearer and avoids keeping around settings that are no longer part of the active selection flow.

#### Duplicated and outdated code was removed
A large portion of the value in this PR comes from cleanup. It removes duplicated and no-longer-used code related to media device selection and older camera-handling utilities.

This includes:
- removing source change methods from preview/background publisher flows
- removing older camera helper utilities and related tests
- deleting abstractions that are no longer needed
- reducing local component logic now that shared selection behavior is available

That cleanup reduces maintenance overhead and makes the code easier to understand.

#### Feedback for device selection errors was improved
The feedback layer is also improved by adding translated, user-facing messaging on top of the shared selection flow.

This gives the implementation a cleaner separation of concerns:
- lower-level selection logic handles the actual device change
- the handler layer is responsible for translation and feedback
- components remain focused on interaction and rendering

### Overall
This PR improves the device-selection architecture by consolidating behavior at the lower-level `selectDevice` flow, removing unused media source settings, and simplifying components so they rely on a shared mechanism instead of local implementations. It also cleans up a significant amount of duplicated and outdated code, making the overall structure more consistent and easier to maintain.

frontend/src/components/MeetingRoom/SmallViewportHeader/hooks/useSwitchCameraFacingModeHandler.ts
```diff
@@ -0,0 +1,120 @@
...
+      const fallbackMessage = t('devices.video.camera.switchError', {
+        facingMode: facingModeLabel,
+      });
+
+      handleClientApplicationError(fallbackMessage, error);
+    }
+  });
+
+  return {
+    /**
```
OscarFava, line 113: Remove this comments

OscarFava (dismissed):

OscarFava:
vcr:deploy

OscarFava:
Please fix conflicts.

johnny-quesada-developer:
> Please fix conflicts.

ready. rebased with develop

frontend/src/components/MeetingRoom/InputAudioDevices/InputAudioDevices.tsx
```diff
@@ -32,9 +35,19 @@ const InputAudioDevices = ({ handleToggle }: InputAudioDevicesProps): ReactEleme
...
+  const handleChangeAudioSource = async (deviceId: string) => {
+    try {
+      handleToggle();
+
+      await handleSelectDevice({
+        deviceId,
+        mediaDeviceKind: 'audioinput',
+      });
+    } catch (error) {
+      handleClientApplicationError(makeApplicationErrorMapper()(error));
```
Hossein-Movahed, line 49: The `handleSelectDevice` throws `ApplicationClientError` with `DeviceAccess`/`DevicesTrackUnavailable` type set and a custom `fallbackMessage`. The component's `makeApplicationErrorMapper()(error)` wraps that again, the second `ApplicationClientError` constructor only preserves type when source is `server_error`. Client error types are silently lost.

Fix is one line  :
 
```suggestion
} catch (error) {
  handleClientApplicationError(error);
}
```

frontend/src/components/MeetingRoom/OutputAudioDevices/OutputAudioDevices.tsx
```diff
@@ -40,12 +43,21 @@ const OutputAudioDevices = ({ handleToggle }: OutputAudioDevicesProps): ReactEle
...
-    if (!isSinkIdSupported()) return;
+      if (!isSinkIdSupported()) return;
 
-    await mediaDevices$.actions.selectDevice('audiooutput', deviceId);
+      await handleSelectDevice({
+        deviceId,
+        mediaDeviceKind: 'audiooutput',
+      });
+    } catch (error) {
+      handleClientApplicationError(makeApplicationErrorMapper()(error));
```
Hossein-Movahed, line 59: [Same here](https://github.com/Vonage/vonage-video-react-app/pull/507/changes#r3318567512)

frontend/src/components/MeetingRoom/VideoDevices/VideoDevices.tsx
```diff
@@ -36,9 +39,19 @@ const VideoDevices = ({
...
+  const handleChangeVideoSource = async (deviceId: string) => {
+    try {
+      handleToggle();
+
+      await handleSelectDevice({
+        deviceId,
+        mediaDeviceKind: 'videoinput',
+      });
+    } catch (error) {
+      handleClientApplicationError(makeApplicationErrorMapper()(error));
```
Hossein-Movahed, line 53: [Same here](https://github.com/Vonage/vonage-video-react-app/pull/507/changes#r3318567512)

frontend/src/components/WaitingRoom/MenuDevices/MenuDevices.tsx
```diff
@@ -8,51 +8,49 @@ import cleanAndDedupeDeviceLabels from '@utils/cleanAndDedupeDeviceLabels/cleanA
...
   const processedDevices = useMemo(() => cleanAndDedupeDeviceLabels(devices), [devices]);
+
   const shouldDisplayDevices =
     mediaDeviceKind !== 'audiooutput' || isGetActiveAudioOutputDeviceSupported();
+
   const shouldDisplayEmptyState = shouldDisplayDevices && processedDevices.length === 0;
 
+  const { handleSelectDevice } = useSelectDeviceHandler();
+
+  const applicationErrorMapper = makeApplicationErrorMapper();
```
Hossein-Movahed, line 52: [same here](https://github.com/Vonage/vonage-video-react-app/pull/507/changes#r3318567512)

frontend/src/components/MeetingRoom/SmallViewportHeader/hooks/useSwitchCameraFacingModeHandler.ts
```diff
@@ -0,0 +1,120 @@
+import { attempt, tryCatch } from '@common/execution';
```
Hossein-Movahed, line 1: No test for this file?

frontend/src/components/WaitingRoom/MenuDevices/MenuDevices.tsx
Hossein-Movahed, line 52: `applicationErrorMapper` created on every render in `MenuDevices` no `useMemo`. Move outside component or wrap: 

 
```suggestion
const applicationErrorMapper = useMemo(() => makeApplicationErrorMapper(), []);
```

Hossein-Movahed (changes requested):

frontend/src/components/MeetingRoom/SmallViewportHeader/hooks/useSwitchCameraFacingModeHandler.ts
johnny-quesada-developer: Hi @OscarFava ,

That's the docs for the public part of the hook API:

Image

frontend/src/components/MeetingRoom/InputAudioDevices/InputAudioDevices.tsx
johnny-quesada-developer: Hi @Hossein-Movahed ,

Is not like that, the type is preserved, there is even a test for that:
image

[Test](https://github.com/Vonage/vonage-video-react-app/pull/507/changes#diff-0d67c76035c673bc791d0ec9eaed462e5b6edc44003ef762dece136f788098edR9)

frontend/src/components/MeetingRoom/OutputAudioDevices/OutputAudioDevices.tsx
johnny-quesada-developer: Hi @Hossein-Movahed ,

[The type is preserved, please check this unit test](https://github.com/Vonage/vonage-video-react-app/pull/507/changes#diff-0d67c76035c673bc791d0ec9eaed462e5b6edc44003ef762dece136f788098edR9)

Inside **ApplicationClientError.ts** type is filled like this:
```tsx
   this.type = (() => {
      if (!isErrorLike(src)) return this.fallbackConfig.type ?? this.type;
      return (src as { type?: string }).type ?? this.fallbackConfig.type ?? this.type;
    })();
```

If the **src** has type is preserved, otherwise uses the **fallbackConfig**, and if there was not **fallbackConfig** keeps 'error' since is the generic type defined on the class:

```tsx
  public override type: string = 'error';
```

frontend/src/components/MeetingRoom/VideoDevices/VideoDevices.tsx
johnny-quesada-developer: [same as here](https://github.com/Vonage/vonage-video-react-app/pull/507#discussion_r3322886179)

frontend/src/components/WaitingRoom/MenuDevices/MenuDevices.tsx
johnny-quesada-developer: [same as here](https://github.com/Vonage/vonage-video-react-app/pull/507#discussion_r3322886179)

frontend/src/components/MeetingRoom/SmallViewportHeader/hooks/useSwitchCameraFacingModeHandler.ts
johnny-quesada-developer: [frontend/src/components/MeetingRoom/SmallViewportHeader/hooks/useSwitchCameraFacingModeHandler.spec.ts](https://github.com/Vonage/vonage-video-react-app/pull/507/changes#diff-6b125f6cc8694c2e3ed6a3f95de137f51acdf6d38260d83b19d51f98b6d9fafc)

frontend/src/components/WaitingRoom/MenuDevices/MenuDevices.tsx
johnny-quesada-developer: Hi @Hossein-Movahed,

**applicationErrorMapper** is just a function. Its creation does not require any heavy computation, so there is no benefit in memoizing the callback. Also, **MenuItem** is not a memoized component, so there is no value in memoizing the parameters passed to it.

Hossein-Movahed (dismissed):

OscarFava (dismissed):

rserebrennykov:
vcr:deploy

rserebrennykov (changes requested):
Image

I am still do able to see the whole list of cameras in a dropd-down list

rserebrennykov:
image

After i tried to change between front and rear cameras started seeing broken video

johnny-quesada-developer:
vrc:deploy

johnny-quesada-developer:
> Image
> I am still do able to see the whole list of cameras in a dropd-down list

Hi @rserebrennykov  ,

As we discussed the other day, this PR specifically fixes the Android issue where the native API returns cameras that are not actually usable.

Unfortunately, the list of cameras with the unusual number sequences comes directly from the Android environment. We do not generate those values, so we cannot change them. Also, the device metadata does not provide enough information to reliably distinguish which cameras work and which ones do not.

Because of that, we cannot know ahead of time which cameras are valid. The only reliable improvement we can make in this case is to validate the camera at the moment the user tries to select it. We need to request permissions at that point in order to check whether the selected camera can provide an actual usable video track.

The workaround we added is similar to the behavior in Google Meet. The main difference is that Google Meet allows the user to switch to the unusable camera and then disables the video, while our approach prevents the device change if the target camera is not usable.

This PR only validates whether the selected Android camera can be accessed and whether it provides a usable media track at the moment the user tries to switch to it.

The current validation flow is:

1. Request access to the selected device using `getUserMedia` with the exact `deviceId`.
2. If access fails, we map that throw `Failed to access ${kind} device: ${label}` 
3. If access succeeds, we stop the returned tracks.
4. If stopping or accessing the track fails, we throw `Failed to access ${kind} track for device: ${label}`

Then related to the other two issues we saw during testing, none of these validations were triggered. That means the selected camera passed this validation successfully, and unfortunately the problem is other not introduced in this PR.

So, these two new issues are related to a different part of the camera-switching lifecycle, and even if they happen in the same area of the product, they are not caused by this validation logic.

For this PR, I would suggest keeping the scope focused on the issue where the native API exposes cameras that appear in the device list but cannot actually provide a usable video track.

The other two issues should be investigated in separate tasks:

1. Android camera switch sometimes turns the video off.
2. Android camera switch can leave the UI in an inconsistent state, where avatar letters are visible, the camera button is disabled, but the video tile still appears enabled.

We should definitely investigate these problems, but this PR represent a fix for another use cases that today we are lacking, we probably should not block this PR because they are not regressions introduced by this fix and they are not related to the validation path added here.

What do you think @rserebrennykov ? Also I will like to know your opinion @VZaphod, @OscarFava  ?

These are the validations we added:
image

And as an example here it is the validation you can get in Google Meet due the same problem:
image


Also here is you can see the list of selected devices:
Google Meet:
image

Ours: 
image

OscarFava:
I think we should create differents PRs to fix that in order to avoid blocking current fixes.

OscarFava (approved):

rserebrennykov (approved):
Agree, will fix the issues in a separate tickets

PR #521, opened 2026-06-08
VIDSOL-688: Advanced settings - Enable Observability Features and Real-Time Stats

OscarFava:
#### What is this PR doing?
Implement support for sender stats and connection metrics. Add an Inspector panel in the UI to display real-time statistics, enabling users and developers to monitor call quality and troubleshoot issues. User preferences must persist using browser storage.

Check metris in Advanced settings.

Resolves [VIDSOL-688](https://jira.vonage.com/browse/VIDSOL-688)

#### Checklist
[ ] Resolves a `Known Issue`.


OscarFava:
vcr:deploy

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/formatters.ts
```diff
@@ -0,0 +1,84 @@
...
+export function formatResolution(resolution: Resolution | null): string {
+  if (!resolution?.width || !resolution?.height) {
+    return '–';
+  }
+
+  return `${resolution.width}x${resolution.height}`;
+}
+
+export function formatFrameRate(frameRate: number | null): string {
+  if (!frameRate) {
```
Hossein-Movahed: the `!frameRate` treats 0 as `falsy`,  0 fps will incorrectly display as '–'

```suggestion
  export const formatFrameRate = (frameRate: number | null): string => {
  if (frameRate == null) return '–';
  return `${Math.round(frameRate)} fps`;
};
```

Same applies to `formatBitrate` and `formatPacketLoss` if they use `!value guards`, please audit all formatter guards for the same pattern.

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/useStatisticsInspectorData/useStatisticsInspectorData.ts
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    return normalizeNonNegative(Math.max(...connectionEstimatedBandwidthValues));
+  })();
+  const packetLossRatio = calculatePacketLossRatio({
+    packetsLost: videoTotals.packetsLost,
+    packetsSuccessful: videoTotals.packetsSent,
+  });
+
+  const currentTimestamp = Date.now();
```
Hossein-Movahed: NIT: Bitrate is always null on the first poll because there's no previous byte sample to diff against.

When the stats panel opens, users will see '–' for all bitrate fields for approximately 1 second until the second poll completes and a delta can be calculated. This is technically correct but can feel broken.

Consider either:

Option A : Show a loading skeleton/spinner for bitrate-specific fields until the first valid value is available 
Option B: Use the SDK's own bandwidth estimate field as a temporary stand-in on the first tick until the delta-based bitrate kicks in.
Option C : Document this intentionally with a comment so future devs don't try to "fix" it.

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/useStatisticsInspectorData/useStatisticsInspectorData.ts
```diff
@@ -0,0 +1,422 @@
...
+    void poll();
+
+    return () => {
+      isCancelled = true;
+
+      if (timeoutIdentifier) {
+        clearTimeout(timeoutIdentifier);
+      }
+    };
+  }, [meetingPublisher, previewPublisher, publisherStatisticsEnabled, subscriberWrappers]);
```
Hossein-Movahed: If `subscriberWrappers` is a new array reference on every parent render, the useEffect will rerun on every render cycle, resetting all stats to empty and restarting the polling loop.

This depends on how `subscriberWrappers` is created in `useSessionContext`. If it's not memoized (e.g., derived inline like `const subscriberWrappers = Object.values(subscribers)`), this will cause a polling reset storm.

possible fix: Stabilize the dependency by using a length + ID fingerprint ref, or ensure `subscriberWrappers` is memoized at the context level

Hossein-Movahed (changes requested):
Awesome work 🔥 , just a small comments

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/useStatisticsInspectorData/useStatisticsInspectorData.ts
```diff
@@ -0,0 +1,422 @@
...
+
+type UseStatisticsInspectorDataArgs = {
+  meetingPublisher: Publisher | null | undefined;
+  previewPublisher: Publisher | null | undefined;
+  subscriberWrappers: SubscriberWrapper[];
+  publisherStatisticsEnabled: boolean;
+};
+
+const useStatisticsInspectorData = ({
+  meetingPublisher,
```
johnny-quesada-developer: Hi @OscarFava ,

Instead of providing two publishers (previewPublisher, meetingPublisher) and deciding from which one to extract the statistics, let's make the hook work with just one publisher:

```tsx
const previewPublisherStats = useStatisticsInspectorData({ publisher: previewPublisher });
const meetingPublisher = useStatisticsInspectorData({ publisher: meetingPublisher });

// or in this case
const meetingPublisher = useStatisticsInspectorData({ publisher: meetingPublisher ?? previewPublisher });
```

Once we move the publisher to the core we'll be removing this publisher duplicates:

```tsx
  const { publisher: meetingPublisher } = usePublisherContext();
  const { publisher: previewPublisher } = usePreviewPublisherContext();
```

And there will be just publishers... 

Then, a session could have more than one publisher, so for example, right now when we screen share we have two publishers. 

So, let's keep track of the publishers a session have, and let's render the statistics from there:

```tsx
  const publish = useCallback(async (publisher: Publisher): Promise<void> => {
    if (!vonageVideoClient.current) {
      return;
    }

   // our VonageVideoClient should keep track of how many publishers it has
    await vonageVideoClient.current.publish(publisher);
  }, []);
```

So then you could do something like:
```tsx
{publishers.map((publisher) => <PublisherStatistics publisher={publisher} />)}
```

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/useStatisticsInspectorData/useStatisticsInspectorData.ts
```diff
@@ -0,0 +1,422 @@
...
+  previewPublisher,
+  subscriberWrappers,
+  publisherStatisticsEnabled,
+}: UseStatisticsInspectorDataArgs): StatisticsInspectorData => {
+  const [data, setData] = useState<StatisticsInspectorData>(EMPTY_DATA);
```
johnny-quesada-developer: Hi @OscarFava ,

For all data retrieval specially for this kind which requiere a pulling mechanism and fail tolerance lets use react query, is already installed .

Example:
```tsx
const useArchives = <Selected = Result>({
  queryOptions,
  sessionKey,
  count,
  offset,
}: UseArchivesProps<Selected>) => {
  const videoClient = runtime$.useVideoClient();

  return runtime$.useQuery({
    ...queryOptions,

    queryKey: ['archives', sessionKey, count, offset],

    queryFn: async () => {
      return await videoClient.searchArchives({
        sessionKey,
        count,
        offset,
      });
    },
  });
};
```

In your case for the terminologies you can take the language from the runtime provider.

```tsx
const language = runtime$.useLanguage();
```

Another example

```tsx
const { data, ...rest } = useArchivesBase({
    sessionKey,
    queryOptions: {
      enabled: !!sessionKey,
      refetchInterval: (query) => {
        const archives = query.state.data?.items;

        if (!archives || !hasPending(archives)) {
          return false;
        }

        return poolingIntervalMs;
      },
    },
  });
```

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/formatters.ts
OscarFava: Should be frameRate === null, thx

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/useStatisticsInspectorData/useStatisticsInspectorData.ts
OscarFava: Option C looks good.

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/useStatisticsInspectorData/useStatisticsInspectorData.ts
johnny-quesada-developer: Hi @Hossein-Movahed,

This feedback looks generated and misaligned. 

This comment is based on an assumption that you can evacuate by looking at the code. The `subscriberWrappers` is memoized in the session provider...

Also, replacing the dependency with a `length` is not reliable. If one subscriber disconnects and another one connects, the length remains the same.

If `subscriberWrappers` changes, then recomputing is expected.

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/useStatisticsInspectorData/useStatisticsInspectorData.ts
```diff
@@ -0,0 +1,422 @@
...
+  const previousPublisherVideoSampleRef = useRef<{ bytesSent: number; timestamp: number } | null>(
+    null
+  );
+
+  useEffect(() => {
```
johnny-quesada-developer: Hi @OscarFava ,

So, we can avoid managing data through a `useEffect` and use react query instead like mentioned [here](https://github.com/Vonage/vonage-video-react-app/pull/521/changes#r3389678232).

Also let's simplify this a little by creating smaller pieces... This hook as it is right now is doing too much at once.

Let's make the logic more atomic so will be easier to asimilate:

```tsx
const {data,error,isLoading} = usePublisherStats({ publisher });
const {data,error,isLoading} = useSubscriberStats({ subscriber });
```

And subsequently

```tsx
{publishers.map((publisher) => <PublisherStats publisher={publisher} />)}
{subscribers.map((subscriber) => <SubscriberStats subscriber={subscriber} />)}
```

Internally they can keep using **AdvancedSettingsStatisticsGroup**...

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/useStatisticsInspectorData/useStatisticsInspectorData.ts
```diff
@@ -0,0 +1,422 @@
...
+  bytesReceived?: number;
+}): IncomingTrackTotals {
+  return {
+    packetsReceived: track?.packetsReceived ?? 0,
+    packetsLost: track?.packetsLost ?? 0,
+    bytesReceived: track?.bytesReceived ?? 0,
+  };
+}
+
+function isNonNullable<T>(value: T | null): value is T {
```
johnny-quesada-developer: Hi @OscarFava ,

We already have `@common/asserts/isNotNil`

johnny-quesada-developer (changes requested):

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/useStatisticsInspectorData/useStatisticsInspectorData.ts
OscarFava: good catch

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/useStatisticsInspectorData/useStatisticsInspectorData.ts
OscarFava: Sure, let's do that. thx

frontend/src/components/AdvancedSettingsDialog/AdvancedSettingsStatisticsTab/useStatisticsInspectorData/useStatisticsInspectorData.ts
OscarFava: Same as above comment, thx

libs/core/src/hooks/usePublisherStats/usePublisherStats.spec.ts
```diff
@@ -0,0 +1,257 @@
+import { describe, expect, it, vi, beforeEach, afterEach, type Mock } from 'vitest';
+import { renderHook } from '@testing-library/react';
+import type { Publisher, PublisherStatsArr } from '@vonage/client-sdk-video';
+import usePublisherStats from './usePublisherStats';
+
+vi.mock('@core/stores', async () => {
```
johnny-quesada-developer: Hi @OscarFava,

I think this test is not very valuable in its current form because it is overmocking the hook internals instead of testing the actual behavior of `usePublisherStats`.

The dependency that needs to be mocked is **`publisher.getStats`**, because that is the async source of the stats data. Both **`useStableRef`** and **`useQuery`** should be allowed to run normally during the test.

Mocking `useQuery` and `useStableRef` makes the test depend on implementation details, so the test can pass even if the real hook behavior is broken, or fail after a valid refactor that does not change the public behavior.

libs/core/src/hooks/usePublisherStats/usePublisherStats.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+  previousSample: { bytesSent: number; timestamp: number } | null = null
+) {
+  let capturedQueryFn: (() => Promise<unknown>) | undefined;
+
+  (runtime$.useQuery as Mock).mockImplementation((options: { queryFn: () => Promise<unknown> }) => {
+    capturedQueryFn = options.queryFn;
+    return { data: undefined, isLoading: true };
+  });
+
+  (useStableRef as Mock).mockReturnValue({ current: previousSample });
```
johnny-quesada-developer:
```suggestion
vi.moked(useStableRef).mockReturnValue({ current: previousSample });
```

johnny-quesada-developer (changes requested):

johnny-quesada-developer (dismissed):
LGTm

johnny-quesada-developer (dismissed):

Hossein-Movahed (dismissed):

rserebrennykov:
vcr:deploy

OscarFava:
vcr:deploy

manolovn (approved):

johnny-quesada-developer (approved):

PR #670, opened 2026-07-07
VIDSOL-1069: fix(recording): don't treat backdrop click / Escape as a recording-consent decline

### LINKED ISSUE #653: Bug: recording-consent dialog treats backdrop click / Escape as "decline" and ejects the participant from the call
## Summary

The recording-consent dialog wires the "decline" action to the dialog's `onClose`. Because MUI `Dialog.onClose` also fires on **backdrop click** and **Escape**, dismissing the dialog the way people dismiss any dialog is interpreted as an explicit *"I do not consent to being recorded"* — and the participant is immediately navigated to `/goodbye` and disconnected from the meeting, with no confirmation.

## Where

`frontend/src/components/MeetingRoom/RecordingPopupIndicator/RecordingPopupIndicator.tsx:47-63`

```tsx
const handleDecline = () => {
  setHasBeenNotified(true);
  redirectToGoodbye();          // navigates to /goodbye and leaves the meeting
};
...
return (
  <PopupDialog
    isOpen={shouldOpenDialog}
    handleClose={handleDecline}   // ← decline wired to onClose
    handleActionClick={handleActionClick}
    actionText={actionText}
  />
);
```

`frontend/src/components/MeetingRoom/PopupDialog/PopupDialog.tsx:42-53`

```tsx
<Dialog open={isOpen} onClose={handleClose} ...>   // fires on backdropClick + escapeKeyDown
  ...
  <Button onClick={handleClose}>{actionText.secondaryActionText}</Button>  // explicit Decline
```

## Mechanism

When any participant starts recording, every *other* participant is shown the consent dialog. MUI calls `onClose(event, reason)` for `reason === 'backdropClick'` and `reason === 'escapeKeyDown'` — the `reason` is never inspected. So a stray click outside the dialog, or pressing Escape, runs `handleDecline` → `redirectToGoodbye()` → the user is ejected from the call as if they had clicked **Decline**.

(`ArchivingButton` reuses the same `PopupDialog` safely, because there `handleClose` only closes the modal — it doesn't leave the meeting.)

## Impact

- Participants are **accidentally kicked out mid-meeting** whenever anyone records, via the single most common dialog-dismissal gesture.
- A legally significant **consent** action is triggered by an ambiguous gesture rather than a deliberate choice.

## How to fix

Consent is mandatory (Accept or Decline) — the dialog should **not** treat backdrop/Escape as an answer. Add an opt-in `disableDismiss` to the shared `PopupDialog` so we don't change behaviour for the archive/mute dialogs:

```diff
// PopupDialog.tsx
 export type DialogProps = {
   isOpen: boolean;
   handleClose: () => void;
   handleActionClick: () => void;
   actionText: DialogTexts;
+  disableDismiss?: boolean;
 };

-const PopupDialog = ({ isOpen, handleClose, handleActionClick, actionText }: DialogProps) => {
+const PopupDialog = ({ isOpen, handleClose, handleActionClick, actionText, disableDismiss = false }: DialogProps) => {
   return (
     <Dialog
       open={isOpen}
-      onClose={handleClose}
+      onClose={disableDismiss ? undefined : handleClose}
+      disableEscapeKeyDown={disableDismiss}
       ...
```

```diff
// RecordingPopupIndicator.tsx
   <PopupDialog
     isOpen={shouldOpenDialog}
     handleClose={handleDecline}
     handleActionClick={handleActionClick}
     actionText={actionText}
+    disableDismiss
   />
```

With `onClose` undefined + `disableEscapeKeyDown`, backdrop click and Escape do nothing; the explicit **Decline** button (which still calls `handleClose`) remains the only way to decline. A spec asserting a backdrop click does **not** navigate to `/goodbye` would lock this in.

---

czoli1976:
## Summary

Fixes #653.

The recording-consent dialog wired **decline** (which navigates to `/goodbye` and leaves the meeting) to the dialog's `onClose`. MUI fires `onClose` for **backdrop click** and **Escape** — the most common way anyone dismisses a dialog — so those gestures **ejected the participant from the call** as if they'd explicitly declined recording. This hits every non-recording participant whenever anyone starts recording.

## Fix

Add an opt-in `disableDismiss` to the shared `PopupDialog` that ignores the implicit dismiss gestures (`backdropClick` / `escapeKeyDown`), and enable it on the consent prompt:

```diff
// PopupDialog.tsx
-onClose={handleClose}
+onClose={(_event, reason) => {
+  const isDismissGesture = reason === 'backdropClick' || reason === 'escapeKeyDown';
+  if (disableDismiss && isDismissGesture) return;
+  handleClose();
+}}
```
```diff
// RecordingPopupIndicator.tsx
 <PopupDialog … actionText={actionText} + disableDismiss />
```

Explicit **Accept**/**Decline** buttons are unaffected (they call their handlers directly). Other `PopupDialog` usages (archive start/stop, mute) are unchanged — `disableDismiss` defaults to `false`, so they keep dismiss-to-close.

Added a regression test to `RecordingPopupIndicator.spec.tsx` **before** the fix:

- **Before** (on `develop`): firing Escape and clicking the backdrop → `expected "spy" to not be called at all, but actually been called 2 times` — both gestures eject the user.
- **After**: neither redirects; `mockNavigate` is not called. The existing "redirects when the user declines" (explicit button) test still passes.

## Testing

- `RecordingPopupIndicator.spec.tsx` + `PopupDialog.spec.tsx`: all pass (new test failed before the fix)
- `yarn test` (full suite) ✅ — 192 test files / 968 passed, 2 skipped

### DIFF
--- a/frontend/src/components/MeetingRoom/PopupDialog/PopupDialog.tsx
+++ b/frontend/src/components/MeetingRoom/PopupDialog/PopupDialog.tsx
@@ -18,11 +18,6 @@ export type DialogProps = {
   handleClose: () => void;
   handleActionClick: () => void;
   actionText: DialogTexts;
-  /**
-   * When true, the dialog cannot be dismissed via backdrop click or Escape;
-   * only the explicit action buttons close it. Use for mandatory prompts (e.g.
-   * recording consent) where an accidental dismissal must not count as a choice.
-   */
   disableDismiss?: boolean;
 };
 
@@ -49,8 +44,6 @@ const PopupDialog = ({
     <Dialog
       open={isOpen}
       onClose={(_event, reason) => {
-        // For mandatory prompts, ignore the implicit dismiss gestures (backdrop click / Escape)
-        // so an accidental dismissal is not treated as an explicit choice.
         const isDismissGesture = reason === 'backdropClick' || reason === 'escapeKeyDown';
         if (disableDismiss && isDismissGesture) {
           return;
--- a/frontend/src/components/MeetingRoom/PopupDialog/PopupDialog.tsx
+++ b/frontend/src/components/MeetingRoom/PopupDialog/PopupDialog.tsx
@@ -18,6 +18,12 @@ export type DialogProps = {
   handleClose: () => void;
   handleActionClick: () => void;
   actionText: DialogTexts;
+  /**
+   * When true, the dialog cannot be dismissed via backdrop click or Escape;
+   * only the explicit action buttons close it. Use for mandatory prompts (e.g.
+   * recording consent) where an accidental dismissal must not count as a choice.
+   */
+  disableDismiss?: boolean;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


VZaphod:
vcr:deploy

VZaphod (approved):
LGTM

johnny-quesada-developer (approved):

PR #672, opened 2026-07-07
VIDSOL-1067: fix(token): set client-token expireTime in epoch seconds, not milliseconds (tokens never expire)

### LINKED ISSUE #637: Bug: client tokens never expire — expireTime passed in milliseconds but the SDK expects seconds (security)
## Summary

Both server-side token paths build `expireTime` as **`Date.now() + <milliseconds>`**, but the `@vonage/video` SDK treats `ClientTokenOptions.expireTime` as an **absolute UNIX time in seconds** and writes it straight into the JWT `exp` claim. The millisecond value (~`1.75e12`) becomes an `exp` in the far future (~year 57,000), so **the generated client tokens effectively never expire.**

## Where

- `libs/api/src/handlers/createEphemeralToken.ts:22` — `expireTime: Date.now() + 30 * 1000` (intended **30s** server-to-server MODERATOR token).
- `backend/routes/video/constants/joinSession.ts:11` — `expireTime: Date.now() + threeHoursInMilliseconds` (intended **3h** publisher join token).

## Why it's wrong

`@vonage/video`'s `generateClientToken` (`node_modules/@vonage/video/dist/cjs/index.js`):

```js
generateClientToken(sessionId, tokenOptions) {
  const now = Math.round(new Date().getTime() / 1e3);   // seconds
  const claims = {
    ...
    exp: tokenOptions?.expireTime || now + 86400,        // exp is in SECONDS, used directly
    ...
```

and the type doc is explicit:

```ts
// @vonage/video ClientTokenOptions.d.ts
/** The expiration time of the client token in seconds (optional). */
expireTime?: number;
```

Two problems compound:
1. **Wrong unit** — `Date.now()` is **ms**, the SDK wants **seconds**.
2. **Wrong semantics** — `exp` is set to `expireTime` *directly* (not `now + expireTime`), so it must be an absolute epoch-seconds timestamp, not a duration.

Either way the result is an `exp` far in the future → tokens that never expire.

## Impact

- **Security:** long-lived/ever-valid client tokens (including the MODERATOR ephemeral token) defeat the short-TTL intent. A leaked token stays valid indefinitely.
- The carefully chosen "30s" / "3h" comments are silently ignored.

## The codebase already does it correctly elsewhere

`backend/videoService/opentokVideoService.ts:103` uses seconds:

```ts
const expires = Math.floor(new Date().getTime() / 1000) + 24 * 60 * 60;
```

## How to fix

Pass an absolute epoch in **seconds** (the SDK writes `expireTime` directly into `exp`):

```diff
// libs/api/src/handlers/createEphemeralToken.ts:22
-      expireTime: Date.now() + 30 * 1000,
+      expireTime: Math.floor(Date.now() / 1000) + 30,
```

```diff
// backend/routes/video/constants/joinSession.ts
-const threeHoursInMilliseconds = 3 * 60 * 60 * 1000;
+const threeHoursInSeconds = 3 * 60 * 60;
 ...
-      expireTime: Date.now() + threeHoursInMilliseconds,
+      expireTime: Math.floor(Date.now() / 1000) + threeHoursInSeconds,
```

Add a unit test asserting the resulting `exp` is within a few seconds of `now + ttl` (and not ~`1e12`).

---
*Observed on `develop` @ `56ae12e`. The router strips a client-supplied `expireTime`, so these server defaults are the values that actually ship.*

cc @OscarFava @VZaphod for triage / routing.

czoli1976:
## Summary

Fixes #637 — a security bug: generated client tokens **never expire**.

`@vonage/video`'s `generateClientToken` writes `ClientTokenOptions.expireTime` **straight into the JWT `exp` claim**, which is an absolute UNIX time in **seconds**. Both token paths passed `Date.now() + <milliseconds>` (~`1.7e12`), so `exp` landed around the year **57000** — the tokens, including the server-to-server **MODERATOR** ephemeral token, effectively never expired, defeating the intended 30s / 3h TTLs.

## Fix

Pass an absolute epoch in **seconds** in both sites:

```diff
// libs/api/src/handlers/createEphemeralToken.ts
-      expireTime: Date.now() + 30 * 1000,
+      expireTime: Math.floor(Date.now() / 1000) + 30,
```
```diff
// backend/routes/video/constants/joinSession.ts
-const threeHoursInMilliseconds = 3 * 60 * 60 * 1000;
+const threeHoursInSeconds = 3 * 60 * 60;
-      expireTime: Date.now() + threeHoursInMilliseconds,
+      expireTime: Math.floor(Date.now() / 1000) + threeHoursInSeconds,
```

This matches how the codebase already computes token expiry elsewhere (`opentokVideoService.ts` uses `Math.floor(new Date().getTime() / 1000) + …`).

Regression tests at both sites assert the `expireTime` is within a few seconds of `now + ttl` **and** `< 1e11` (seconds, not milliseconds):

- **Before** (on `develop`): both fail — `expected 1781645355997 to be less than 5` (the ms / far-future value).
- **After**: both pass.

## Testing

- `createEphemeralToken.test.ts` (libs/api, vitest) ✅ · `joinSession.test.ts` (backend, jest) ✅ — both failed before the fix
- `yarn test` (full suite) ✅ — frontend 192 files / 967, backend 21 suites / 119, libs/api 28 files / 88

### DIFF
--- a/backend/routes/video/constants/joinSession.test.ts
+++ b/backend/routes/video/constants/joinSession.test.ts
@@ -1,0 +1,18 @@
+import { describe, expect, it } from '@jest/globals';
+import joinSession from './joinSession';
+
+describe('joinSession defaults', () => {
+  it('sets the client token expireTime in epoch seconds (~3h from now), not milliseconds', () => {
+    const nowSeconds = Math.floor(Date.now() / 1000);
+
+    const addDefaults = joinSession.addDefaults as (payload: unknown) => {
+      clientTokenOptions: { expireTime: number };
+    };
+    const { expireTime } = addDefaults({}).clientTokenOptions;
+
+    // The SDK writes expireTime straight into the JWT `exp` (epoch seconds), so it
+    // must be ~now + 3h in seconds. The bug used Date.now() + milliseconds (~1.7e12).
+    expect(Math.abs(expireTime - (nowSeconds + 3 * 60 * 60))).toBeLessThan(5);
+    expect(expireTime).toBeLessThan(1e11);
+  });
+});
--- a/backend/routes/video/constants/joinSession.ts
+++ b/backend/routes/video/constants/joinSession.ts
@@ -1,14 +1,15 @@
 import type { HandlersConfig } from '@api-lib';
 import { TokenRole } from '@api-lib';
 
-const threeHoursInMilliseconds = 3 * 60 * 60 * 1000;
+const threeHoursInSeconds = 3 * 60 * 60;
 
 const joinSession: HandlersConfig['joinSession'] = {
   addDefaults: (payload) => ({
     ...payload,
     clientTokenOptions: {
       role: TokenRole.MODERATOR,
-      expireTime: Date.now() + threeHoursInMilliseconds,
+      // expireTime is an absolute UNIX time in seconds (the SDK writes it into the JWT `exp`).
+      expireTime: Math.floor(Date.now() / 1000) + threeHoursInSeconds,
       ...payload.clientTokenOptions,
     },
   }),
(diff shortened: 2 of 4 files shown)


johnny-quesada-developer (approved):
LGTM

OscarFava (approved):

VZaphod (approved):
LGTM

rserebrennykov (approved):
lgtm

PR #674, opened 2026-07-07
VIDSOL-1065: fix(storage): make InMemorySessionStorage.setSession idempotent (recordings never auto-stopped)

### LINKED ISSUE #654: Bug: recordings & captions are never auto-stopped — setSession wipes tracked archiveIds on every join (cost/privacy)
## Summary

`InMemorySessionStorage.setSession` **unconditionally re-creates** the room's session record (resetting `archiveIds: []`, `captionsId: null`, `captionsUserCount: 0`), and it is called on **every `joinSession`**, not just on session creation. So the moment any participant joins (or refreshes) after a recording starts, the tracked `archiveIds` are wiped — and the `sessionDestroyed` webhook can no longer stop the archive. **Recordings (and captions) keep running and billing after the room empties.**

## Where

`backend/storage/inMemorySessionStorage/inMemorySessionStorage.ts:20-39`

```ts
async setSession({ roomName, sessionKey, sessionId }): Promise<void> {
  this.sessions[roomName] = {
    sessionKey,
    captionsId: null,
    captionsUserCount: 0,
    archiveIds: [],          // ← blows away accumulated archive ids / captions state
  };
  this.roomNameBySessionKey[sessionKey] = roomName;
  if (sessionId) this.sessionKeyBySessionId[sessionId] = sessionKey;
}
```

`backend/routes/video/video.ts:55-67` — the middleware runs `setSession` for **join**, not only create:

```ts
videoHandler.onSettled$(async ({ videoAction, error, result }) => {
  const isCreatingOrJoiningSession =
    !error &&
    [VideoAction.createSession, VideoAction.createSessionAndJoin, VideoAction.joinSession]
      .includes(videoAction);
  if (!isCreatingOrJoiningSession) return;
  const { sessionId, sessionKey, roomName } = result as VideoSessionDetails;
  await sessionService.setSession({ sessionId, sessionKey, roomName: roomName! });
});
```

## Mechanism

1. User A creates the room and starts recording → the `/hooks/archive` webhook accumulates the id via `setArchiveIds` → `archiveIds: ['ARCHIVE_1']`.
2. User B joins → `onSettled$` fires for `joinSession` → `setSession` **re-creates** the record → `archiveIds: []`.
3. Everyone leaves → the `/hooks/session` `sessionDestroyed` handler reads `getArchiveIds()` → `[]` → never calls `stopArchive`. The recording runs until it hits Vonage's max duration / storage limits.

The same wipe resets `captionsId`/`captionsUserCount`, desyncing the captions refcount. (VCR's `VcrSessionStorage` stores archiveIds under a separate key and is unaffected — so the two storage backends have divergent semantics for the same interface; this hits local/self-hosted deployments.)

## Impact

- **Cost / privacy:** archives (and captions) continue after the meeting ends — the exact outcome the `sessionDestroyed` handler comment says it exists to prevent ("essential to avoid unnecessary costs").
- Captions user-count desync.

## How to fix

Make `setSession` **idempotent** — preserve existing per-room state when the record already exists, only initializing on first creation, while always refreshing the key/id mappings:

```diff
 async setSession({ roomName, sessionKey, sessionId }): Promise<void> {
-  this.sessions[roomName] = {
-    sessionKey,
-    captionsId: null,
-    captionsUserCount: 0,
-    archiveIds: [],
-  };
+  const existing = this.sessions[roomName];
+  this.sessions[roomName] = existing
+    ? { ...existing, sessionKey }
+    : { sessionKey, captionsId: null, captionsUserCount: 0, archiveIds: [] };
   this.roomNameBySessionKey[sessionKey] = roomName;
   if (sessionId) this.sessionKeyBySessionId[sessionId] = sessionKey;
 }
```

(Alternatively — or additionally — restrict the `onSettled$` middleware to the create actions only, but the storage-level idempotency fix is the safer guarantee.)

---

czoli1976:
## Summary

Fixes #654.

`InMemorySessionStorage.setSession` unconditionally re-created the room record (`archiveIds: []`, `captionsId: null`, `captionsUserCount: 0`), and it runs on **every** `joinSession` (via the `onSettled$` middleware). So any participant joining **after** a recording started wiped the tracked `archiveIds`. The `sessionDestroyed` hook then read `[]` and never called `stopArchive` (or `disableCaptions`) — so **recordings and captions kept running and billing after the room emptied**, the exact outcome that hook exists to prevent.

## Fix

Make `setSession` idempotent — preserve existing per-room state when the record already exists, only initializing on first creation, while always refreshing the key/id mappings:

```diff
-this.sessions[roomName] = { sessionKey, captionsId: null, captionsUserCount: 0, archiveIds: [] };
+const existingSession = this.sessions[roomName];
+this.sessions[roomName] = existingSession
+  ? { ...existingSession, sessionKey }
+  : { sessionKey, captionsId: null, captionsUserCount: 0, archiveIds: [] };
```

Added a regression test: set a session, record an archive id + captions id, then call `setSession` again for the same room (simulating a later join).

- **Before** (on `develop`): `getArchiveIds` returns `[]` — `expected [] to equal ['archive-1']` — the ids were wiped.
- **After**: `archiveIds` and `captionsId` survive the repeat `setSession`.

## Testing

- `inMemorySessionStorage.test.ts`: 20 suites pass, incl. the new idempotency test (which failed before the fix)
- `yarn test` (full suite) ✅ — backend 20 suites / 114, frontend 967, libs/api 88

Fixes #654.

### DIFF
--- a/backend/storage/inMemorySessionStorage/inMemorySessionStorage.test.ts
+++ b/backend/storage/inMemorySessionStorage/inMemorySessionStorage.test.ts
@@ -104,4 +104,20 @@ describe('InMemorySessionStorage', () => {
       );
     });
   });
+
+  describe('setSession idempotency', () => {
+    it('preserves archiveIds and captionsId when setSession is called again for the same room', async () => {
+      await storage.setSession({ roomName: room, sessionKey, sessionId });
+      await storage.setArchiveIds({ sessionId, archiveIds: ['archive-1'] });
+      await storage.setCaptionsId({ sessionId, captionsId: 'captions-1' });
+
+      // A later participant joining re-triggers setSession for the same room; it must
+      // not wipe the tracked archives/captions, which still need to be stopped on
+      // sessionDestroyed.
+      await storage.setSession({ roomName: room, sessionKey, sessionId });
+
+      expect(await storage.getArchiveIds({ sessionId })).toEqual(['archive-1']);
+      expect(await storage.getCaptionsId({ sessionId })).toBe('captions-1');
+    });
+  });
 });
--- a/backend/storage/inMemorySessionStorage/inMemorySessionStorage.ts
+++ b/backend/storage/inMemorySessionStorage/inMemorySessionStorage.ts
@@ -26,12 +26,19 @@ class InMemorySessionStorage implements SessionStorage {
     sessionKey: string;
     sessionId: string;
   }): Promise<void> {
-    this.sessions[roomName] = {
-      sessionKey,
-      captionsId: null,
-      captionsUserCount: 0,
-      archiveIds: [],
-    };
+    // Idempotent: setSession runs on every join, so preserve existing per-room
+    // state (archiveIds / captions) when the record already exists. Re-creating it
+    // would wipe tracked archives and captions, so sessionDestroyed could no longer
+    // stop them.
+    const existingSession = this.sessions[roomName];
+    this.sessions[roomName] = existingSession
+      ? { ...existingSession, sessionKey }
+      : {
+          sessionKey,
+          captionsId: null,
+          captionsUserCount: 0,
+          archiveIds: [],
+        };
     this.roomNameBySessionKey[sessionKey] = roomName;
     if (sessionId) {
       this.sessionKeyBySessionId[sessionId] = sessionKey;


czoli1976:
@copilot Review this PR Carefully:

- Is this issue still present on Develop today?
- Apply your improvement suggestions only if very high confidence (no need if small)
- Check if the CI is fine, if not fix the issues
- Provide a Summary of the Value of this PR for @VZaphod with a reccomendation for him regarding keep/drop/merge and priority

PR #675, opened 2026-07-07
VIDSOL-1064: fix(background): clear the effect when deleting the applied custom image

### LINKED ISSUE #645: Bug: deleting the applied custom background re-applies a background instead of clearing it
## Summary

When you delete a **custom background image that is currently applied**, the code's comment says it should *clear* the background — but it instead **re-applies a background** (`backgroundSelected`), so the effect is not removed; it silently switches to whatever `backgroundSelected` holds.

## Where

`frontend/src/Context/BackgroundPublisherProvider/useBackgroundPublisher/useBackgroundPublisher.tsx:253-258`:

```ts
// If the deleted image was the currently applied background filter, clear it
const currentBackgroundFilter = getInitialBackgroundFilter(backgroundPublisherRef.current);
if (imageToDelete.dataUrl === currentBackgroundFilter) {
  changeBackground(backgroundSelected).catch(() => {        // re-applies, does not clear
    throw new Error('Failed to reset background filter after deleting custom image');
  });
}
```

## Mechanism

The intent (per the comment) is to remove the now-deleted background. Instead it calls `changeBackground(backgroundSelected)`, which **applies** `backgroundSelected`. If `backgroundSelected` still refers to the just-deleted image (or to a different effect than the live one), the user does not get the background cleared — they get a stale/other effect re-applied.

## Impact

Deleting the active custom background does not turn the effect off as the UI implies; the preview can jump to a different background instead of returning to "no effect".

## How to fix

Apply the explicit `'none'` option (the value the codebase already uses to clear a background — see `useBackgroundPublisher.spec.tsx`, `changeBackground('none')`) instead of re-applying `backgroundSelected`:

```diff
// useBackgroundPublisher.tsx:255-258
   if (imageToDelete.dataUrl === currentBackgroundFilter) {
-    changeBackground(backgroundSelected).catch(() => {
+    changeBackground('none').catch(() => {
       throw new Error('Failed to reset background filter after deleting custom image');
     });
   }
```

---
*Observed on `develop` @ `56ae12e`.*

cc @OscarFava @VZaphod for triage / routing.

czoli1976:
## Summary

Fixes #645.

When you delete a **custom background image that is currently applied**, `deleteCustomImage`'s "clear it" branch called `changeBackground(backgroundSelected)` — which **re-applies** the currently-selected background instead of clearing the effect. So deleting an applied-but-unselected custom image silently switched to a different effect rather than turning it off.

## Fix

```diff
 if (imageToDelete.dataUrl === currentBackgroundFilter) {
-  changeBackground(backgroundSelected).catch(() => {
+  changeBackground('none').catch(() => {
     throw new Error('Failed to reset background filter after deleting custom image');
   });
 }
```

`'none'` is the value the codebase already uses to clear a background (see `applyBackgroundFilter` and the `changeBackground('none')` spec).

Added a regression test: seed a custom image as the *applied* filter while a *different* background is selected, then delete it.

- **Before** (on `develop`): `expected "spy" to be called at least once` — `clearVideoFilter` is never called; the delete re-applies the selected background instead.
- **After**: the filter is cleared (`clearVideoFilter` called; no `backgroundReplacement` re-applied).

## Testing

- `useBackgroundPublisher.spec.tsx`: 11/11 pass (the new test failed before the fix)
- `yarn test` (full suite) ✅ — frontend 192 files / 968, backend 21 suites, libs/api 28 files

### DIFF
--- a/frontend/src/Context/BackgroundPublisherProvider/useBackgroundPublisher/useBackgroundPublisher.spec.tsx
+++ b/frontend/src/Context/BackgroundPublisherProvider/useBackgroundPublisher/useBackgroundPublisher.spec.tsx
@@ -160,6 +160,50 @@ describe('useBackgroundPublisher', () => {
     });
   });
 
+  describe('deleteCustomImage', () => {
+    it('clears the background when the applied custom image is deleted, instead of re-applying the selected one', async () => {
+      const dataUrl = 'data:image/png;base64,AAAA';
+
+      const publisher = Object.assign(new EventEmitter(), {
+        getAudioSource: () => defaultAudioDevice,
+        getVideoSource: () => defaultVideoDevice,
+        applyVideoFilter: vi.fn(),
+        clearVideoFilter: vi.fn(),
+        // The currently-applied filter IS the custom image we are about to delete.
+        getVideoFilter: () => ({ type: 'backgroundReplacement', backgroundImgUrl: dataUrl }),
+      }) as unknown as Publisher;
+
+      mockedHasMediaProcessorSupport.mockReturnValue(true);
+      mockedInitPublisher.mockReturnValue(publisher);
+
+      const { result } = render({
+        hookInitialValue: {
+          customImages: [{ id: 'img1', dataUrl }],
+          // A different background is selected; the bug would re-apply this instead of clearing.
+          backgroundSelected: 'bg1.jpg',
+        },
+      });
+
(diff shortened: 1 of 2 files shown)


VZaphod (approved):
LGTM

VZaphod:
vcr:deploy

OscarFava (approved):

PR #698, opened 2026-07-08
VIDSOL-1043: fix(hooks): disconnect ResizeObserver on unmount in useElementDimensions (#564)

### LINKED ISSUE #564: Bug: useElementDimensions ResizeObserver never disconnected (memory leak)
## Problem

In `frontend/src/hooks/useElementDimensions.tsx`, the `ResizeObserver` is never properly disconnected:

```tsx
const resizeObserver = useRef<ResizeObserver | undefined>(undefined);

useEffect(() => {
  const elementCurrent = elementRef.current;
  if (elementCurrent && !resizeObserver.current) {
    resizeObserver.current = new ResizeObserver(() => {
      throttledSetDimensions();
    });
    resizeObserver.current?.observe(elementCurrent);
  }
  return () => {
    if (elementCurrent) {
      resizeObserver.current?.unobserve(elementCurrent); // ← only unobserves
    }
    // ← never calls .disconnect()
    // ← never clears the ref
  };
}, [elementRef]);
```

### Issues:
1. `.unobserve()` removes one target but the observer itself remains in memory
2. `resizeObserver.current` is never set back to `undefined`, so on re-mount the guard `!resizeObserver.current` prevents creating a new observer — the ref holds a disconnected/stale observer
3. The `throttledSetDimensions` closure captures the original `elementCurrent`, potentially holding a reference to a detached DOM element

## Fix

```tsx
useEffect(() => {
  const elementCurrent = elementRef.current;
  if (!elementCurrent) return;

  const throttledSetDimensions = throttle(() => {
    setElementDimensions({
      height: elementCurrent.offsetHeight,
      width: elementCurrent.offsetWidth,
    });
  }, 20);

  const observer = new ResizeObserver(() => {
    throttledSetDimensions();
  });
  observer.observe(elementCurrent);

  return () => {
    observer.disconnect();
  };
}, [elementRef]);
```

## Impact

- Memory leak: observers accumulate over component lifecycle
- Stale observer prevents resize tracking after re-mount
- Particularly relevant for VideoTileCanvas which uses this hook for the main video grid container

czoli1976:
## Summary

Fixes #564. `useElementDimensions`'s effect cleanup called `resizeObserver.current?.unobserve(elementCurrent)` but **never `disconnect()`ed the observer and never reset the ref**. Two consequences:

- **Leak:** the `ResizeObserver` was never released — it accrues per mount of any component using the hook (video tiles, panels, etc.).
- **Stale observer:** because `resizeObserver.current` stayed set, the `!resizeObserver.current` guard blocked re-creating an observer if `elementRef` changed, so the hook silently stopped tracking size after a ref swap.

## Fix

Cleanup now disconnects and drops the observer — fixing both issues:

```diff
     return () => {
-      if (elementCurrent) {
-        resizeObserver.current?.unobserve(elementCurrent);
-      }
+      resizeObserver.current?.disconnect();
+      resizeObserver.current = undefined;
     };
```

No behavioral/visual change while mounted — `disconnect()` covers the single observed element that `unobserve` handled, and additionally releases the observer and re-arms the guard.

## Testing

Test-first: the spec stubs `ResizeObserver` and asserts `disconnect()` is called once on unmount — **fails on develop** (0 calls; only `unobserve` ran) and passes with the fix.

- Full suite green (`yarn test`); `yarn quality-check` clean.

> Scoped to the reported leak: the shared `throttle` helper (no public `.cancel()`) is left untouched, so a pending trailing `setState` after unmount is out of scope (React 19 no-ops it).


czoli1976:
@copilot review carefully this PR and

-  Confirm if this issue is still present on  Develop today
-  Rebase if needed so we do not have conflicts as Develop moved
-  Apply all High Confidence Improvements 
-  Assuming this is still needed then Explain Value to @VZaphod and provide a recommendation on what to do next

PR #703, opened 2026-07-11
VIDSOL-1081: fix(notifications): stabilize onDismiss so item auto-dismiss timers aren't reset (#557)

### LINKED ISSUE #557: Bug: NotificationsContainer handleDismiss resets auto-dismiss timers on every re-render
## Problem

In `libs/ui/src/components/NotificationsContainer/NotificationsContainer.tsx`, `handleDismiss` is defined as a plain inline function:

```tsx
const handleDismiss = (id: NotificationId) => {
  notifications$.actions.dismiss(id);
};
```

This creates a new function reference on every render. Since `NotificationItem` has a `useEffect` that depends on `onDismiss`:

```tsx
useEffect(() => {
  if (expirationMs === null) return;
  const timer = setTimeout(() => onDismiss(id), expirationMs);
  return () => clearTimeout(timer);
}, [id, expirationMs, onDismiss]);
```

Every time the parent re-renders, **all notification auto-dismiss timers are cleared and restarted**. This means notifications may never auto-dismiss if the parent re-renders frequently enough.

## Fix

Wrap `handleDismiss` with `useCallback`:

```tsx
const handleDismiss = useCallback((id: NotificationId) => {
  notifications$.actions.dismiss(id);
}, []);
```

Since `notifications$.actions.dismiss` is a stable store action, the dependency array is empty and the function reference stays stable across renders.

## Impact

- Notifications reliably auto-dismiss after their configured `expirationMs`
- Eliminates unnecessary `useEffect` cleanup/re-setup cycles in every `NotificationItem`

czoli1976:
## Summary

Fixes #557. `NotificationsContainer` passed an **inline** `handleDismiss` (a new function reference on every render) to each `NotificationItem`. `NotificationItem`'s auto-dismiss effect keys on `[id, expirationMs, onDismiss]`:

```ts
useEffect(() => {
  if (expirationMs === null) return;
  const timer = setTimeout(() => onDismiss(id), expirationMs);
  return () => clearTimeout(timer);
}, [id, expirationMs, onDismiss]);
```

So every time the container re-renders — e.g. when a **second** notification arrives — `onDismiss`'s identity changes, the effect is torn down and re-created, and each item's `setTimeout` restarts. Under a steady stream of notifications an item's countdown can be perpetually reset and dismiss late (or never).

## Fix

Wrap `handleDismiss` in `useCallback` so its reference is stable. One line, `libs/ui`, no visual change.

## Testing

Test-first: `NotificationsContainer.spec` asserts every rendered `NotificationItem` receives the **same** `onDismiss` reference across a container re-render — fails on `develop` (2 distinct references), passes with the fix.

- Full suite green (`yarn test`); `yarn quality-check` clean.


czoli1976:
@copilot  review this PR very carefully and

- Determine if this is still needed
- Apply the highest-confidence recommended changes you may have and make sure the CI is green
- Make sure no confits / rebase needed, if so take care of it right away
- Assuming it is still needed then explain very clearly the value of this PR and tag @VZaphod so He can take a decision on what to do next with this PR

czoli1976:
@copilot chase @VZaphod for a follow

PR #709, opened 2026-07-12
VIDSOL-1075: fix(devices): kill the Firefox device-store readiness deadlock (#657, Facet A)

### LINKED ISSUE #657: Bug: Firefox device store — isStoreReady can deadlock or stay rejected, breaking device sync & selection
## Summary

On Firefox, the media-device store's readiness promise `meta.isStoreReady` can either **deadlock (never resolve)** or **stay permanently rejected**, which kills all subsequent device syncing and device selection for the session. Two distinct facets, same root (the `isStoreReady` lifecycle in `setupDeviceStore`).

## Facet A — circular-await deadlock (single-use flag consumed by a concurrent `getUserMedia`)

`libs/core/src/stores/mediaDevices/helpers/getMediaDevicesInfo/getMediaDevicesInfo.ts:14-29`

```ts
const shouldSkipStoreReady = metadata.isFirstMediaDevicesInfoQuery;
metadata.isFirstMediaDevicesInfoQuery = false;      // single-use global flag
...
if (!shouldSkipStoreReady) {
  await metadata.isStoreReady;                       // ← can await the very promise it must resolve
}
```

`libs/core/src/stores/mediaDevices/helpers/setupDeviceStore/setupDeviceStore.ts:44-74` (Firefox branch requests permission, then resolves via `syncMediaDevicesInfo`):

```ts
meta.isStoreReady = new CancelablePromise((resolve, reject, { isCanceled }) => {
  const syncDevicesAndResolve = () =>
    void api.actions.syncMediaDevicesInfo().then(() => resolve()).catch(reject);
  if (!isFirefox()) { void syncDevicesAndResolve(); return; }
  void navigator.mediaDevices.enumerateDevices()
    .then((devices) => {
      if (devices.some((d) => d.label)) return;
      return getUserMedia({ audio: true, video: true }).then((s) => s.getTracks().forEach((t) => t.stop())); // permission prompt — long window
    })
    .then(() => syncDevicesAndResolve())
    .catch(reject);
});
```

`libs/core/src/stores/mediaDevices/actions/getUserMedia/getUserMedia.ts:13-16` — every monkey-patched `getUserMedia` (including the Vonage SDK publisher acquiring media in the waiting room) re-enters the sync:

```ts
return _getUserMedia(constraints).then((stream) => {
  void this.syncMediaDevicesInfo();     // ← consumes isFirstMediaDevicesInfoQuery
  return stream;
});
```

**Mechanism:** On Firefox first visit, `isStoreReady`'s executor is parked on the `getUserMedia({audio,video})` permission prompt. During that window the publisher's own `getUserMedia` (proxied to the store) resolves and calls `syncMediaDevicesInfo` → `getMediaDevicesInfo` → **consumes the `isFirstMediaDevicesInfoQuery` flag** (sets it `false`). When the bootstrap prompt then resolves, `syncDevicesAndResolve` → `syncMediaDevicesInfo` → `getMediaDevicesInfo` now sees the flag `false` → `await metadata.isStoreReady` — the promise it is trying to resolve. **Circular wait; `isStoreReady` stays pending forever.** (Non-Firefox is safe only because the flag is consumed synchronously inside the executor before any concurrent call.)

## Facet B — permanent rejection poisons the store

`setupDeviceStore.ts:38-42`

```ts
const syncMediaDevicesInfoDebounced = debounce(async () => {
  await meta.isStoreReady;                    // rejects forever after a one-time failure
  void api.actions.syncMediaDevicesInfo().catch(() => {});
}, 10);
```

**Mechanism:** The Firefox bootstrap `getUserMedia({audio,video})` rejects **atomically** if the user denies the prompt, **or** has a mic but **no camera** (Firefox fails combined constraints with `NotFoundError` — see the special-case in `getUserMedia.ts:22-24`). That rejection flows to `.catch(reject)` (`setupDeviceStore.ts:73`) and `isStoreReady` is **never replaced**. Thereafter every `devicechange` / permission-change / `visibilitychange` handler awaits a rejected promise: the `await meta.isStoreReady` at line 39 throws **before** the `.catch(() => {})` at line 41, so each event produces an **unhandled promise rejection** and never syncs. Even the permission-change listener (lines 81-95) that fires exactly when the user later *grants* access can't recover the store.

## Impact (Firefox)

- **Deadlock (Facet A):** `selectDevice` awaits `meta.loadingMediaDevices`/`isStoreReady` → switching camera/mic hangs forever; unplug-detection dead; the whole device layer freezes after at most one successful sync.
- **Rejection (Facet B):** mic-only Firefox machines (no webcam — very common) and any deny-then-grant flow leave device menus empty/unlabeled for the whole session; console floods with unhandled rejections on every device/visibility change.

## How to fix

1. **Kill the deadlock — make "skip readiness" explicit, not a shared consumable flag.** Have `getMediaDevicesInfo` accept a `skipStoreReady` argument (the bootstrap `syncDevicesAndResolve` passes `true`), instead of reading/mutating `metadata.isFirstMediaDevicesInfoQuery`, so a concurrent `getUserMedia`-driven sync can't consume the bootstrap's skip:
   ```ts
   getMediaDevicesInfo({ skipStoreReady = false } = {}) { return async ({ getMetadata }) => {
     if (!skipStoreReady) await getMetadata().isStoreReady;
     ... } }
   ```
2. **Don't reject `isStoreReady` on a permission problem — resolve it.** Device labels being empty is a normal pre-permission state, not a fatal store error; resolving lets later permission-grant resyncs work:
   ```diff
   -    .then(() => syncDevicesAndResolve())
   -    .catch(reject);
   +    .then(() => syncDevicesAndResolve())
   +    .catch(() => resolve());   // proceed unlabeled; a later permission-change will resync
   ```
   and request audio/video **independently** (`Promise.allSettled`) so a mic-only machine isn't failed atomically.
3. **Guard the debounced await** so a (transiently) rejected store can't emit unhandled rejections:
   ```diff
   -  await meta.isStoreReady;
   -  void api.actions.syncMediaDevicesInfo().catch(() => {});
   +  await meta.isStoreReady.catch(() => {});
   +  void api.actions.syncMediaDevicesInfo().catch(() => {});
   ```

---
*Observed on `develop` @ `1ac3822`. Both facets are code-verified; runtime reproduction requires Firefox (deadlock needs the publisher's `getUserMedia` to land during the bootstrap permission window; rejection needs deny, or a mic-only device).*

czoli1976:
## Summary

Partially fixes #657 — **Facet A (the circular-await deadlock)**.

`getMediaDevicesInfo` decided whether to `await metadata.isStoreReady` by reading and clearing a **single-use** metadata flag (`isFirstMediaDevicesInfoQuery`). On Firefox's first visit the bootstrap sync is parked on the `getUserMedia` permission prompt while the publisher's own `getUserMedia` (proxied to the store) resolves and drives a **concurrent** sync that consumes the flag. When the prompt then resolves, the bootstrap sync sees the flag `false` and `await`s `isStoreReady` — **the very promise it is responsible for resolving**. Circular wait: `isStoreReady` never resolves and the whole device layer (device sync, `selectDevice`, unplug detection) freezes for the session.

## Fix

Make "skip readiness" an **explicit argument** instead of a shared consumable flag, so a concurrent `getUserMedia`-driven sync can't steal the bootstrap's skip:

- `getMediaDevicesInfo({ skipStoreReady })` → `syncMediaDevicesInfo({ skipStoreReady })`, with `setupDeviceStore`'s bootstrap passing `skipStoreReady: true`.
- Removed the now-unused `isFirstMediaDevicesInfoQuery` metadata flag.

Pure-logic change; no behavior change for non-bootstrap callers (default `skipStoreReady: false`).

## Testing

Test-first, and **deterministically reproducible without Firefox**: `getMediaDevicesInfo.spec` sets exactly the deadlock state — `isStoreReady` pending and the first-query flag already consumed — and asserts the bootstrap sync (`skipStoreReady: true`) resolves within a timeout. **Fails on `develop`** (awaits the pending promise forever), passes with the fix.

- Full `core` suite green (89 tests); `yarn quality-check` clean. One existing Firefox `setupDeviceStore` test updated for the new call signature.

## Deliberately out of scope: Facet B

The issue's **Facet B** (`isStoreReady` staying *rejected* on a Firefox permission deny / mic-only device) requires a permission-flow change — resolve-instead-of-reject **and** requesting audio/video independently (`Promise.allSettled`) — whose runtime behavior (e.g. whether Firefox shows one prompt or two) **can't be verified in CI/without a real Firefox**. Rather than ship an unverified change to the permission flow, I've left Facet B as a follow-up. Happy to add it with clear "needs Firefox QA" framing if preferred.


czoli1976:
@copilot  review this PR carefuly and

- Determine if this is still needed
- Apply the recommended changes you may have, only the highest-confidence ones
- Make sure no confits / rebase needed; if so take care of it
- Assuming it is still needed, then explain very clearly the value of this PR and tag @VZaphod so He can take a decision on what to do next with this PR

PR #714, opened 2026-07-15
VIDSOL-1112: fix(publisher): surface publishing errors after a sustained transient state, not just once

### LINKED ISSUE #713: Bug: publish-failure safety net never fires if the session stays 'transient' forever, leaving users stuck unpublished
## Summary

A user reported joining a meeting and being unable to be seen or heard, while he could see and hear everyone else. Client console/network logs from the incident show:

```
OT.exception :: title: Unable to Publish (1500) msg: ICEWorkflow   (x3, at +0ms / +51s / +25s later)
OpenTok:RaptorSocket:error cannot publish until the socket is connected.
Subscriber with stream ID ... destroyed   (x6, in a burst)
/client-logs: Failed to load resource: 502
```

This is an **asymmetric-connectivity network issue** on the client (his network allowed subscribing but blocked the WebRTC publish path) — not fixable in application code. However, digging into `usePublisher.tsx`, the app already has a safety net built for exactly this scenario, and it has a gap that let this user fall through it silently.

## Root cause

`frontend/src/Context/PublisherProvider/usePublisher/usePublisher.tsx` listens for `OT.on('exception', ...)`. After **3 consecutive** non-transient `1500` (Unable to Publish) failures, it sets `publishingError`, which `useMeetingRoom.ts`'s `useRedirectOnPublisherError` uses to redirect the user to `/goodbye` — its own comment says: *"If the user is unable to publish, we redirect them... This prevents users from subscribing to other participants in the room, and being unable to communicate with them."*

```ts
const shouldTreatAsTransient = reconnectingRef.current || !connected || !isBrowserOnline;
...
const shouldSurfaceBlockingError =
  shouldTreatAsTransient === false && consecutivePublishingFailureCountRef.current >= 3;
```

The bug: `shouldTreatAsTransient` treats **any** reconnecting/disconnected/offline state as an indefinite free pass, on the assumption it will "self-heal." The incident's log shows a burst of 6 subscribers being destroyed right around the 3rd failure — strong evidence the signaling connection (`connected`) dropped at that exact moment. If `connected` is `false` when the 3rd exception lands, `shouldTreatAsTransient` is `true`, the redirect is suppressed, and — because no further `1500` exception fires once retries stop — **the safety net never re-evaluates**. The user is left stuck in the room indefinitely: publishing silently broken, no error shown, no redirect.

## How to fix

Track how long a failure streak has persisted (`firstPublishingFailureAtRef`, reset on a successful publish). If failures keep occurring past a grace period even while nominally "transient," surface the error anyway — a transient outage that never resolves is not actually transient:

```diff
+ const failureDurationMs = Date.now() - firstPublishingFailureAtRef.current;
+ const hasExceededTransientGracePeriod =
+   failureDurationMs >= PUBLISHING_TRANSIENT_FAILURE_GRACE_PERIOD_MS; // 45s

  const shouldSurfaceBlockingError =
-   shouldTreatAsTransient === false && consecutivePublishingFailureCountRef.current >= 3;
+   consecutivePublishingFailureCountRef.current >= 3 &&
+   (shouldTreatAsTransient === false || hasExceededTransientGracePeriod);
```

## Reproduction

Test-first: reproduce the incident's exact timing (3 `1500` failures at 0s/+51s/+25s while `connected` stays `false` throughout — mirroring the real log). `publishingError` stays `null` forever on unpatched `develop`; the fix surfaces it once the grace period is exceeded.

---
*Filed from a real production incident triage (client console + network log from the user's session). Fix implemented and test-first verified; PR incoming.*

czoli1976:
## Summary

Fixes #713 — filed from a real production incident. A user reported joining a meeting and being unable to be seen or heard, while he could see and hear everyone else. Client console/network logs from the incident show:

```
OT.exception :: title: Unable to Publish (1500) msg: ICEWorkflow   (×3, at +0ms / +51s / +25s)
OpenTok:RaptorSocket:error cannot publish until the socket is connected.
Subscriber with stream ID ... destroyed   (×6, in a burst)
```

The root network cause (his network blocked the WebRTC publish path while subscribe worked) isn't fixable in application code. But the app already has a safety net designed for exactly this — after 3 consecutive non-transient `1500` failures, `usePublisher` sets `publishingError`, which redirects the user to `/goodbye` rather than leaving them stuck. The incident exposed a gap in it.

## Root cause

```ts
const shouldTreatAsTransient = reconnectingRef.current || !connected || !isBrowserOnline;
...
const shouldSurfaceBlockingError =
  shouldTreatAsTransient === false && consecutivePublishingFailureCountRef.current >= 3;
```

`shouldTreatAsTransient` treats **any** reconnecting/disconnected/offline state as an indefinite free pass, assuming it will self-heal. The incident's log shows a burst of 6 subscribers destroyed right around the 3rd failure — evidence the signaling connection (`connected`) dropped at that exact moment. If `connected` is `false` when the 3rd exception lands, the redirect is suppressed — and since no further `1500` exception fires once retries stop, **the safety net never re-evaluates**. The user is left stuck indefinitely: publishing silently broken, no error, no redirect.

## Fix

Track how long a failure streak has persisted (`firstPublishingFailureAtRef`, reset on a successful publish). If failures keep occurring past a 45s grace period even while nominally "transient," surface the error anyway — a transient outage that never resolves is not actually transient:

```diff
+ const failureDurationMs = Date.now() - firstPublishingFailureAtRef.current;
+ const hasExceededTransientGracePeriod =
+   failureDurationMs >= PUBLISHING_TRANSIENT_FAILURE_GRACE_PERIOD_MS;

  const shouldSurfaceBlockingError =
-   shouldTreatAsTransient === false && consecutivePublishingFailureCountRef.current >= 3;
+   consecutivePublishingFailureCountRef.current >= 3 &&
+   (shouldTreatAsTransient === false || hasExceededTransientGracePeriod);
```

## Testing

Test-first: reproduces the incident's **exact timing** (3 `1500` failures at 0s/+51s/+25s while `connected` stays `false` throughout) — `publishingError` stays `null` forever on unpatched `develop`, is set once the grace period is exceeded with the fix.

- Full suite green (`yarn test`, 193 files / 972 tests); `yarn quality-check` clean.
- Rebased onto latest `develop`.

### DIFF
--- a/frontend/src/Context/PublisherProvider/usePublisher/usePublisher.spec.tsx
+++ b/frontend/src/Context/PublisherProvider/usePublisher/usePublisher.spec.tsx
@@ -1,6 +1,6 @@
-import { beforeEach, describe, it, expect, vi } from 'vitest';
+import { afterEach, beforeEach, describe, it, expect, vi } from 'vitest';
 import { act, renderHook as renderHookBase, waitFor } from '@testing-library/react';
-import {
+import OT, {
   initPublisher,
   Publisher,
   Stream,
@@ -389,6 +389,59 @@ describe('usePublisher', () => {
       });
 
       expect(mockedSessionPublish).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('OT exception handling (sustained network/session failures)', () => {
+    afterEach(() => {
+      vi.useRealTimers();
+    });
+
+    it('surfaces a blocking error after repeated 1500 failures even while the session never fully reconnects', () => {
+      vi.useFakeTimers();
+      mockedInitPublisher.mockReturnValue(mockPublisher);
+
+      // The session stays disconnected throughout, exactly like the real incident where the
+      // signaling connection dropped and never recovered ("connected" never becomes true again).
+      const { result } = renderHook(() => usePublisher(), {
+        sessionContext: {
+          __interceptor: (ctx) => {
+            if (ctx) {
+              ctx.connected = false;
+            }
+          },
+        },
+      });
+
+      act(() => {
+        result.current.initializeLocalPublisher({});
+      });
+
+      const exceptionHandler = vi
+        .mocked(OT.on)
+        .mock.calls.find(([event]) => event === 'exception')?.[1] as (event: {
+        code: number;
+      }) => void;
+      expect(exceptionHandler).toBeDefined();
+
+      expect(result.current.publishingError).toBeNull();
+
+      // 3 failures spanning past the grace period, mirroring the real incident's ~76s span.
+      act(() => {
+        exceptionHandler({ code: 1500 });
+      });
+      act(() => {
+        vi.advanceTimersByTime(51_000);
+        exceptionHandler({ code: 1500 });
+      });
+      act(() => {
+        vi.advanceTimersByTime(25_000);
+        exceptionHandler({ code: 1500 });
+      });
+
+      // Even though the session never stopped looking "transient" (still disconnected), the
+      // user must eventually be told rather than left stuck silently and indefinitely.
+      expect(result.current.publishingError).not.toBeNull();
     });
   });
 
--- a/frontend/src/Context/PublisherProvider/usePublisher/usePublisher.tsx
+++ b/frontend/src/Context/PublisherProvider/usePublisher/usePublisher.tsx
@@ -31,6 +31,14 @@ type DeviceAccessStatus = {
   camera: boolean | undefined;
 };
 
+/**
+ * How long (in ms) a sustained streak of publish failures is tolerated while the session appears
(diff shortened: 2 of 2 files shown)


czoli1976:
@VZaphod - this is a tentative fix but please think about what the end behviour should actually be so that I can change this PR if needed

czoli1976:
@copilot Review this PR Carefully and

- Check if this Issue is still present on Develop, if not then ABORT
- If this does not look like a good way to fix it then do not be shy, say so, it will not hurt my feelings

Assuming the issue is still there and the solution seems sound then, and only then
- Apply your highest confidence improvements
- Make sure no conflicts are present; if there are then address them

Remember to not make unrelated changes which are stricly necessarely, do not add narrative comments anywhere

Issue #617, opened 2026-06-30
Bug: console flooded with "getLayout apply" — stray debug log in opentok-layout-js dependency


czoli1976:
## Summary

The browser console prints **`getLayout apply`** repeatedly during a call. It's a **stray `console.log` left in the third-party `opentok-layout-js` dependency**, not in VERA's own code — but VERA's layout code triggers it on every relayout, so it floods the console.

## Root cause

`opentok-layout-js` ships a debug log inside its `getLayout` wrapper:

```js
// opentok-layout-js  (src/index.ts, and the built opentok-layout.js / .min.js)
getLayout: (...args) => {
  console.log('getLayout apply');           // <-- stray debug log
  return getLayout.apply(this, [opts, ...args]);
},
```

- Installed version: **`opentok-layout-js@5.5.0`** (our dep is `"opentok-layout-js": "^5.5.0"`).
- The log is present in the library's **upstream `master`** too — so **bumping the version won't remove it**.

## Why it's so noisy in VERA

VERA re-creates the layout manager and calls `getLayout` on **every** layout pass:

```ts
// frontend/src/utils/layoutManager.ts
getLayout(containerDimensions, elements, shouldMakeLargeTilesLandscape): Box[] {
  this.init(containerDimensions, shouldMakeLargeTilesLandscape); // re-creates the manager every call
  return this.manager?.getLayout(elements)?.boxes ?? [];
}
```

(The re-create-every-time is itself a workaround for [opentok-layout-js#141](https://github.com/aullman/opentok-layout-js/issues/141).) So every tile add/remove, resize, or active-speaker change fires the log.

## Suggested fix

**Interim (in this repo) — `patch-package`:** strip the line from the installed package so it stops immediately, until upstream PR #144 merges and releases.

1. `npm i -D patch-package` and add a `postinstall` script: `"postinstall": "patch-package"`.
2. Remove `console.log('getLayout apply');` from the consumed build file(s) in `node_modules/opentok-layout-js/` (`opentok-layout.js` and `opentok-layout.min.js`).
3. `npx patch-package opentok-layout-js` and commit the generated `patches/opentok-layout-js+5.5.0.patch`.

Resulting patch (shape):

```diff
--- a/node_modules/opentok-layout-js/opentok-layout.js
+++ b/node_modules/opentok-layout-js/opentok-layout.js
@@ getLayout wrapper
       getLayout: function () {
-        console.log('getLayout apply');
[...]

PR #718, opened 2026-07-19
fix(frontend): memoize hasMediaProcessorSupport to stop deprecation console flood

### LINKED ISSUE #717: Bug: console flooded with 'OT.hasMediaProcessorSupport is deprecated' — deprecated sync API called every render
When testing SDK 2.35 RC

### Summary

The browser console is continuously flooded with:

```
OT.hasMediaProcessorSupport is deprecated, and will be removed in the future. Use OT.hasMediaProcessorSupport.promise instead +1ms
```

This is the same class of problem as #617 (console flooded by a dependency's log), but a different source: it comes from `@vonage/client-sdk-video`, driven by how VERA calls the API.

### Root cause

The SDK's **synchronous** `hasMediaProcessorSupport()` is deprecated and logs a deprecation warning on **every call** (it is the first line of the function). VERA calls this deprecated sync API **directly in component render bodies**, so every React re-render re-invokes it. During a call, re-renders are frequent (network-quality polling, audio levels, layout/participant changes), so the warning is emitted continuously.

It is not only log noise: on Safari the sync check builds an `OffscreenCanvas` and calls `getContext('webgpu')` on **every** call — real wasted work on a value that is constant for the life of the page.

Render-path call sites (fire on every re-render):

- `frontend/src/components/WaitingRoom/MenuMoreOptions/MenuMoreOptions.tsx`
- `frontend/src/components/MeetingRoom/DeviceSettingsMenu/DeviceSettingsMenu.tsx`
- `frontend/src/components/MeetingRoom/ScreenshareVideoTile/ScreenshareVideoTile.tsx`
- `frontend/src/components/MeetingRoom/ReduceNoiseTestSpeakers/ReduceNoiseTestSpeakers.tsx`
- `frontend/src/components/WaitingRoom/BackgroundEffects/BackgroundEffectsButton/BackgroundEffectsButton.tsx`

Cooler call sites (inside callbacks/effects — occasional, not the flood): `usePublisherOptions.tsx`, `useBackgroundPublisher.tsx`, `usePreviewPublisher.tsx`, `applyBackgroundFilter.ts`. There is also one ref-callback call in `libs/ui` `VividIcon.tsx` (infrequent).

### Proposed fix

Media-processor support depends only on the browser, so the result is constant for the page. Add a memoized wrapper that calls the SDK once per media type and route callers through it — removing both the warning spam and the repeated work:

```ts
// frontend/src/utils/hasMediaProcessorSupport/hasMediaProcessorSupport.ts
import { hasMediaProcessorSupport as vonageHasMediaProcessorSupport } from '@vonage/client-sdk-video';

type MediaProcessorMediaType = NonNullable<Parameters<typeof vonageHasMediaProcessorSupport>[0]>;
const supportCache = new Map<MediaProcessorMediaType, boolean>();

const hasMediaProcessorSupport = (mediaType: MediaProcessorMediaType = 'both'): boolean => {
  const cached = supportCache.get(mediaType);
  if (cached !== undefined) return cached;
  const isSupported = vonageHasMediaProcessorSupport(mediaType);
  supportCache.set(mediaType, isSupported);
  return isSupported;
};

export default hasMediaProcessorSupport;
```

Swap the frontend imports from `@vonage/client-sdk-video` to this wrapper. (`VividIcon` in `libs/ui` renders rarely and is a ref-callback, not a flood source — it can be migrated in a follow-up if a shared home is preferred.)

### Notes

Independent of the device-permission work. Related: #617 (same symptom, different dependency).

czoli1976:
## What

Stops the continuous console flood of:

```
OT.hasMediaProcessorSupport is deprecated, and will be removed in the future. Use OT.hasMediaProcessorSupport.promise instead
```

Closes #717.

## Why

The SDK's synchronous `hasMediaProcessorSupport()` is deprecated and logs a deprecation warning on **every call**. Several components call it **directly in their render body**, so it re-runs on every re-render — flooding the console during a call. On Safari the sync check also rebuilds an `OffscreenCanvas`/WebGPU context on every call, so it is wasted work too. Media-processor support is constant for the life of the page, so it only needs to be computed once.

## How

- Add `frontend/src/utils/hasMediaProcessorSupport/` — a memoized wrapper that calls the SDK once per media type and caches the result.
- Route the frontend callers through the wrapper instead of importing `hasMediaProcessorSupport` from `@vonage/client-sdk-video` directly.
- Because the value is now memoized, component tests that stub the SDK check with different values across cases clear the memo via `resetMediaProcessorSupportCache()` in `beforeEach` (exported for that purpose).

## Scope / follow-ups

- `libs/ui` `VividIcon.tsx` still calls the SDK directly (a ref-callback, not a render-path flood source). Left as an optional follow-up since sharing the wrapper across `libs` is a separate placement decision.

## Testing

- New unit test for the wrapper (memoization + per-type caching).
- Full frontend suite green.
- Verified in-browser: the deprecation warning no longer floods during a call.

## Notes

Independent of the device-permission work. Related: #617 (same "console flooded by a dependency log" symptom, different source).

### DIFF
--- a/frontend/src/Context/BackgroundPublisherProvider/useBackgroundPublisher/useBackgroundPublisher.tsx
+++ b/frontend/src/Context/BackgroundPublisherProvider/useBackgroundPublisher/useBackgroundPublisher.tsx
@@ -4,9 +4,9 @@ import {
   Event,
   initPublisher,
   VideoFilter,
-  hasMediaProcessorSupport,
   PublisherProperties,
 } from '@vonage/client-sdk-video';
+import hasMediaProcessorSupport from '@utils/hasMediaProcessorSupport/hasMediaProcessorSupport';
 import usePermissions from '../../../hooks/usePermissions';
 import useUserContext from '../../../hooks/useUserContext';
 import { DEVICE_ACCESS_STATUS } from '../../../utils/constants';
--- a/frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.spec.tsx
+++ b/frontend/src/Context/PreviewPublisherProvider/usePreviewPublisher/usePreviewPublisher.spec.tsx
@@ -5,6 +5,7 @@ import EventEmitter from 'node:events';
 import { defaultAudioDevice, defaultVideoDevice } from '@utils/mockData/device';
 import { DEVICE_ACCESS_STATUS } from '@utils/constants';
 import usePreviewPublisher from './usePreviewPublisher';
+import { resetMediaProcessorSupportCache } from '@utils/hasMediaProcessorSupport/hasMediaProcessorSupport';
 import { makeTestProvider, providers, type ProviderOptions } from '@test/providers';
 import renderAsyncHook from '@web-test/renderAsyncHook';
 import composeProviders from '@web/helpers/composeProviders';
@@ -24,6 +25,7 @@ describe('usePreviewPublisher', () => {
   const mockedHasMediaProcessorSupport = vi.fn();
 
   beforeEach(() => {
+    resetMediaProcessorSupportCache();
     vi.spyOn(console, 'error').mockImplementation(vi.fn());
 
     setupWindowNavigatorMock({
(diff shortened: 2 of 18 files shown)


czoli1976:
Closing to reopen from the main repo branch so E2E tests can run (fork-based PRs are excluded from E2E by ). The branch  already exists in  and will be used for the replacement PR.

Issue #723, opened 2026-07-21
Bug: Firefox lights the camera LED on the landing page — device-label workaround (#373) runs at global init instead of when required


czoli1976:
### Describe the bug

On **Firefox**, the camera indicator LED turns on immediately on the **landing page** (the create/join-a-room page), before entering the Waiting Room and before the user does anything that requires a camera.

This is a side effect of the intentional Firefox device-label workaround added in #373 (`VIDSOL-488: firefox-video-preview`). Firefox returns empty `MediaDeviceInfo.label`s until a media permission has been granted for the origin, so to populate the device list / video preview the store proactively calls `getUserMedia({ audio: true, video: true })`, reads the now-labelled devices, and immediately stops the tracks.

The workaround itself is reasonable — the problem is **when and how** it runs:

1. It runs from the global `mediaDevicesStore` `onInit` callback (`libs/core/src/stores/mediaDevices/mediaDevicesStore.ts` → `setupDeviceStore`), which initializes at app startup — i.e. on the **landing page**, long before the Waiting Room preview that actually needs the labels.
2. It requests **`video: true`**, which is what lights the camera LED (audio alone would not).

Even though the tracks are stopped immediately, acquiring the stream lights the camera LED, so the camera appears active on the landing page when the app has no need for it yet.

Relevant code — `libs/core/src/stores/mediaDevices/helpers/setupDeviceStore/setupDeviceStore.ts`:

```ts
if (!isFirefox()) {
  void syncDevicesAndResolve();
  return;
}

void navigator.mediaDevices
  .enumerateDevices()
  .then((devices) => {
    if (isCanceled()) return;
    const hasLabels = devices.some((device) => device.label);
    if (hasLabels) return;
    // we should request permissions to be able to see the devices labels.
    return getUserMedia({ audio: true, video: true }).then((stream) => {
      stream.getTracks().forEach((track) => track.stop());
    });
  })
  .then(() => syncDevicesAndResolve())
  .catch(reject);
```

### Describe the expected behavior

The camera should only be acquired when it is actually required (e.g. when the Waiting Room preview / device selection is shown), not on the landing page. On the landing page the camera LED should stay off, matching Chrome/Safari (which never take this eager path — it is gated behind `isFirefox()`).

### Include steps to reproduce

Steps to reproduce:
1. Open the app in Firefox with camera/mic permission **not** yet granted for the origin (or set to "Always Ask").
[...]

PR #373, opened 2026-02-17
VIDSOL-488: firefox-video-preview

johnny-quesada-developer:
#### What is this PR doing?
This PR addresses Firefox-specific issues with video preview by filtering out devices with empty deviceIds before permission is granted, and implements proper persistence of audio/video enabled states across publisher initialization. The changes also introduce monkey patching of navigator.mediaDevices.getUserMedia to keep the device store synchronized when getUserMedia is called outside the store's actions.

Changes:

Added filtering of devices with empty deviceIds to handle Firefox's behavior before permissions are granted
Implemented monkey patching of navigator.mediaDevices.getUserMedia to automatically sync device store state
Changed publisher initialization to read audio/video enabled states from localStorage instead of always resetting to enabled
Updated schema validation to reject MediaDeviceInfo objects with empty deviceIds

#### What are the relevant tickets?
A maintainer will add this ticket number.

Resolves [VIDSOL-488](https://jira.vonage.com/browse/VIDSOL-488)

#### Checklist
[ ] Branch is based on `develop` (not `main`).
[ ] Resolves a `Known Issue`.
[ ] If yes, did you remove the item from the `docs/KNOWN_ISSUES.md`? 
[ ] Resolves an item reported in `Issues`.
If yes, which issue? [Issue Number](https://github.com/Vonage/vonage-video-react-app/issues/)?

PR #724, opened 2026-07-21
VIDSOL-1111: fix(devices): defer Firefox device-label permission request off the landing page

czoli1976:
#### What is this PR doing?

Fixes the Firefox-only bug where the **camera LED lights up on the landing page** (before entering the Waiting Room). See #723.

Firefox hides `MediaDeviceInfo.label`s until a media permission has been granted for the origin. To populate the device-selection UI, #373 (`VIDSOL-488: firefox-video-preview`) added a proactive `getUserMedia({ audio: true, video: true })` in `setupDeviceStore`. That call runs from the global `mediaDevicesStore` `onInit` — i.e. at app startup, on the landing page — so on Firefox it acquires the camera (lighting the LED) long before the camera is actually needed.

This PR keeps the label workaround but **defers it to when it is actually required**:

- Removes the eager `getUserMedia` from `setupDeviceStore`'s init path. The store-ready promise now resolves via a plain device sync for all browsers, so the landing page never acquires media.
- Adds a new `requestDeviceLabels` action (`libs/core/.../actions/requestDeviceLabels`) that performs the Firefox enumerate → `getUserMedia` → stop-tracks dance. It is a no-op on non-Firefox and when labels are already present.
- Calls `requestDeviceLabels()` on **Waiting Room mount** (`useWaitingRoom`), where a camera preview is already shown and media is legitimately required.

Chrome/Safari behaviour is unchanged (they never took the eager path — it was gated behind `isFirefox()`).

**Note for reviewers:** on Firefox, `requestDeviceLabels` requests `video: true` so the video-device dropdown shows real camera names even when the user's camera preference is off. It stops the tracks immediately, so this is a split-second camera acquisition in the Waiting Room, not a held-open camera — verified on `meet.vonagenetworks.net` (camera off → brief LED flash on entry, then off). This matches the app's existing Waiting Room behaviour and only moves the label request off the landing page, so no further change is needed here.

On **Firefox**, with camera/mic permission not yet granted for the origin (or set to "Always Ask"):

1. Load the landing page (create/join a room). **Do not** enter a room.
   - ✅ Expected: the camera hardware LED stays **off** (previously it turned on here).
2. Enter a room → Waiting Room.
   - ✅ Expected: permission is requested here; the camera preview shows and device-selection dropdowns list real device names.
3. Repeat on **Chrome/Safari** to confirm no behaviour change (LED off on landing, preview + labels in the Waiting Room).

Automated coverage:
- `setupDeviceStore.test.ts` now asserts media is **not** acquired at init, even on Firefox with empty labels.
- New `requestDeviceLabels.spec.ts` covers the Firefox (empty vs. present labels) and non-Firefox (no-op) paths.
- Full `core`, `frontend`, and `backend` suites pass.

#### Checklist
- [x] Branch is based on `develop` (not `main`).
- [ ] Resolves a `Known Issue`.
- [ ] If yes, did you remove the item from the `docs/KNOWN_ISSUES.md`?
- [x] Resolves an item reported in `Issues`.
If yes, which issue? [#723](https://github.com/Vonage/vonage-video-react-app/issues/723)


czoli1976:
@copilot  review this PR, rebase if neeeded, explain Value

czoli1976:
@copilot can you determine where this problem originated from, which PR so we know who touched this code and you can tag them to look into this

czoli1976:
> > @copilot can you determine where this problem originated from, which PR so we know who touched this code and you can tag them to look into this
> 
> I traced it to PR #373 (`VIDSOL-488: firefox-video-preview`), merged as `c08a52d` on 2026-02-20. That PR introduced the Firefox-specific eager `enumerateDevices()` → `getUserMedia({ audio: true, video: true })` path in `libs/core/src/stores/devices/helpers/setupDeviceStore/setupDeviceStore.ts` (the file was later moved under `stores/mediaDevices`).
> 
> The commits in that PR touching this area were `6c8599d`, `7d25d03`, and `eae98a1`, all by @johnny-quesada-developer, so that is the right person to tag for follow-up.

@copilot ask him to review this then and explain the rationale behind that change

johnny-quesada-developer:
Hi @czoli1976 ,

The solution is a lot simpler, you have to make lazy the import of the:

`import { mediadevices$ } from '@core/stores';`

Right now the barrel file `@core/stores` includes `mediadevices$`, and we import the stores on the App.tsx, which means that we trigger the permissions request.

The solution is to remove `mediadevices$` from the barrel file,  make the dependency lazy for the rooms, so that when you arrive to the landing page there is no request, but once you move to waiting or meeting room we request for the permissions.

You change on the other hand could introduce a lot of complexity and errors because the apps expects the devices to be setup, and also you are adding a synchronization step which could make the reusability of the `mediadevices$` harder instead of self contained like now.

czoli1976:
Hi @johnny-quesada-developer, thanks for looking at this.

I went down the barrel/lazy route first and it doesn't quite get there on its own, because the store is also reached through deep imports that don't go via `@core/stores`: `App.tsx` statically imports `WaitingRoom`/`MeetingRoom` and their providers (nothing in `frontend/src` is `React.lazy` today), and `useWaitingRoom`, `usePreviewPublisher`, `useSyncPublisherDevices`, `useBackgroundPublisher` and `speakingDetector` all import `@core/stores/mediaDevices` directly; `publisher$` → `getFacingMode` pulls it in too. Since `createGlobalState` runs `onInit` in the constructor, the first of those to evaluate still triggers `getUserMedia` on the landing page.

Making it lazy properly means route-level code-splitting for both room trees + their providers and removing `mediaDevices$` and `publisher$` from the barrel — a bigger change, and any future static import on the landing graph would silently reintroduce the bug with nothing to catch it.

That's why I went with keeping the store side-effect-free at init and revealing labels from the one place that needs them; `setupDeviceStore.test.ts` locks that in. `isStoreReady` still resolves via the normal device sync, so consumers see the same state as before — only Firefox labels arrive a moment later, on the page where the preview already requests media.

Happy to go the code-splitting route instead if you'd prefer that direction for the app — just want to make sure it's the size of change you have in mind. Which would you like?

PR #727, opened 2026-07-22
VIDSOL-1110: fix(effects): reuse the preview publisher's camera track for the effects preview (#619)

### LINKED ISSUE #619: Enhancement: Video Effects preview should clone the publisher's track, not re-open the camera (double capture)
## Summary

The Video Effects panel **does** preview on a separate, un-published `BackgroundPublisher` (good) — but that publisher **re-opens the same camera via a device id** instead of **reusing/cloning the main publisher's existing video track**. So while the panel is open, the camera is captured **twice** (two `getUserMedia` opens + two background-filter media-processor pipelines on the same device). This is the double-capture anti-pattern we've advised customers against; VERA should be the reference for the right pattern (clone the existing track).

## Current behavior

`useBackgroundPublisher` builds the preview publisher with a **deviceId** as `videoSource`:

```ts
// frontend/src/Context/BackgroundPublisherProvider/useBackgroundPublisher/useBackgroundPublisher.tsx  (initBackgroundLocalPublisher)
const publisherOptions: PublisherProperties = {
  insertDefaultUI: false,
  videoFilter,
  resolution: env.DEFAULT_RESOLUTION,
  videoSource: mediaDevices$.getState().videoinput,   // <-- deviceId string -> fresh getUserMedia
  publishAudio: false,
  publishVideo: isVideoEnabled,
};
backgroundPublisherRef.current = initPublisher(undefined, publisherOptions, ...);
```

When `videoSource` is a **deviceId**, the SDK opens the device again. The main publisher (`usePublisherOptions` → `videoSource: useDeviceId('videoinput')`) is already capturing that same camera, so opening the effects panel produces a **second concurrent capture of the same device**.

## Why it matters

- **Device contention:** some OS/browser combinations don't allow a second open of the same camera → "camera already in use" / failed preview.
- **Wasted CPU/power:** two capture pipelines **and** two background-filter (media-processor) passes running at once on the same input.
- **Inconsistency:** the two opens can negotiate different exposure/resolution, so the preview doesn't exactly match the live stream.
- **Reference value:** this is precisely the scenario we coach customers on — VERA modelling the wrong approach is a miss.

## Proposed change

Feed the preview publisher a **clone of the existing publisher's track** instead of a deviceId, so only **one** camera capture exists and the preview shares it (`videoSource` accepts a `MediaStreamTrack`):

```ts
// get the active track from the relevant source publisher
// (meeting -> main publisher; waiting room -> preview publisher)
const sourceTrack = sourcePublisher?.getVideoSource()?.track ?? null;

const publisherOptions: PublisherProperties = {
  insertDefaultUI: false,
  videoFilter,
  resolution: env.DEFAULT_RESOLUTION,
  videoSource: sourceTrack ? sourceTrack.clone() : mediaDevices$.getState().videoinput, // clone, fall back to deviceId
  publishAudio: false,
  publishVideo: isVideoEnabled,
};
```

- **Clone, not the same track instance**, so the preview can stop/own its copy independently and the background filter runs on the preview's clone without disturbing the live stream.
- **Lifecycle:** stop the cloned track when the panel closes / the background publisher is destroyed, so we don't leak a track or keep the camera "in use".
- **Fallback:** if no source track is available (e.g. camera currently off), fall back to the deviceId path.
- **Device changes:** when the selected camera changes, re-derive the clone from the new source track (the existing `useSyncPublisherDevices` wiring).

## Affected code

- `frontend/src/Context/BackgroundPublisherProvider/useBackgroundPublisher/useBackgroundPublisher.tsx` — `initBackgroundLocalPublisher` (videoSource), plus track cleanup on teardown
- source of the track: main publisher (`PublisherProvider`) in meeting mode / `PreviewPublisherProvider` in waiting room — the background publisher needs a handle to the active source track
- (reusability) per repo conventions, the "preview off a cloned track" pattern is a good candidate to live as a reusable primitive in `libs/core`

## Notes

- Verified on `develop`: the separate preview publisher exists (since PR #190 / VIDSOL-105), but it uses the deviceId path — hence the double open. This issue is specifically about switching it to a cloned track.

czoli1976:
#### What is this PR doing?

Fixes #619. The Waiting Room **background-effects preview** ran its own publisher (`useBackgroundPublisher`) initialised with `videoSource: <deviceId>`, which opened a **second physical camera capture** alongside the main preview publisher. Observed in the Waiting Room: two concurrent `getUserMedia` camera opens — a **1080p** preview capture and a **720p** effects capture.

This clones the preview publisher's already-captured video track and passes it as the effects publisher's `videoSource`. The Vonage SDK accepts a `MediaStreamTrack` for `videoSource` (`opentok.d.ts`: `videoSource?: string | null | boolean | MediaStreamTrack`), so no second `getUserMedia` is issued. The effects publisher **owns the clone** and stops it on destroy, so the capture is released rather than leaked.

- `useBackgroundPublisher.initBackgroundLocalPublisher(sharedVideoTrack?)` uses the shared track when provided, else falls back to the device id.
- `useWaitingRoom` passes `publisher.getVideoSource().track.clone()` once the preview track is live.
- `destroyBackgroundPublisher` stops the owned clone.

#### Verification (empirical, via Playwright/WebDriver)

- **Physical camera opens: 2 → 1** — confirmed in both **Firefox** and **Chromium** (only the preview's 1080p capture remains).
- **Effects still work on the shared track** — `applyVideoFilter({ type: 'backgroundBlur', blurStrength: 'high' })` on the effects publisher **succeeds** on the cloned track in **Chromium** (`getVideoFilter()` returns the filter, track stays live, no errors). Media processors remain Chromium-only, unchanged.
- **No leak** — the cloned track is `live` in the Waiting Room and `ended` after leaving it.
- Unit tests: added coverage that the shared track is used as `videoSource` and stopped on destroy; full frontend suite green.

1. In **Chrome**, enter a Waiting Room. Open Background Effects and apply a blur/replacement — it should look identical to before.
2. Confirm only **one** camera capture is active (e.g. `chrome://media-internals`, or the OS camera indicator) rather than two.
3. Leave the Waiting Room and confirm the camera is released.

#### Known limitations / follow-ups (not in this PR)

- **Device switching:** the effects publisher's `useSyncPublisherDevices` calls `setVideoSource(deviceId)` on a camera change, which would reopen the camera. On initial mount the clone's deviceId matches the selection, so it does not reacquire; a mid-session camera switch would. Worth a follow-up to re-source the clone from the main publisher instead.
- **Meeting room:** the effects publisher also inits in `useMeetingRoom`; #619 is about the Waiting Room effects *preview*, so this PR scopes there. The same sharing could be extended to the meeting publisher.
- Visual output was verified via the SDK filter APIs (filter applied, track live), not pixel diffing.

Resolves [VIDSOL-](https://jira.vonage.com/browse/VIDSOL-) / Resolves #619

#### Checklist
- [x] Branch is based on `develop` (not `main`).
- [ ] Resolves a `Known Issue`.
- [x] Resolves an item reported in `Issues`. Which issue? [#619](https://github.com/Vonage/vonage-video-react-app/issues/619)


czoli1976:
@copilot review this PR and

- Determine if this is still neeeded
- Apply the reccomended changes you may have
- Mke sure no confits / rebase needed, if so take care of it

Assuming it is still needed then explain the value of this PR and tag @VZaphod

PR #733, opened 2026-07-26
VIDSOL-717: fix(frontend): memoize hasMediaProcessorSupport to stop deprecation console flood

### LINKED ISSUE #717: Bug: console flooded with 'OT.hasMediaProcessorSupport is deprecated' — deprecated sync API called every render

czoli1976:
## What

Stops the continuous console flood of:

```
OT.hasMediaProcessorSupport is deprecated, and will be removed in the future. Use OT.hasMediaProcessorSupport.promise instead
```

Closes #717.

## Why

The SDK's synchronous `hasMediaProcessorSupport()` is deprecated and logs a deprecation warning on **every call**. Several components call it **directly in their render body**, so it re-runs on every re-render — flooding the console during a call. On Safari the sync check also rebuilds an `OffscreenCanvas`/WebGPU context on every call, so it is wasted work too. Media-processor support is constant for the life of the page, so it only needs to be computed once.

## How

- Add `frontend/src/utils/hasMediaProcessorSupport/` — a memoized wrapper that calls the SDK once per media type and caches the result.
- Route the frontend callers through the wrapper instead of importing `hasMediaProcessorSupport` from `@vonage/client-sdk-video` directly.
- Because the value is now memoized, component tests that stub the SDK check with different values across cases clear the memo via `resetMediaProcessorSupportCache()` in `beforeEach` (exported for that purpose).

## Scope / follow-ups

- `libs/ui` `VividIcon.tsx` still calls the SDK directly (a ref-callback, not a render-path flood source). Left as an optional follow-up since sharing the wrapper across `libs` is a separate placement decision.

## Testing

- New unit test for the wrapper (memoization + per-type caching).
- Full frontend suite green.
- Verified in-browser: the deprecation warning no longer floods during a call.

## Notes

Independent of the device-permission work. Related: #617 (same "console flooded by a dependency log" symptom, different source).


czoli1976:
@copilot resolve the merge conflicts in this pull request

czoli1976:
@copilot resolve conflicts

johnny-quesada-developer:
this is not fixing the problem, the correct solution will be to properly migrate to OT.hasMediaProcessorSupport.promise... Since is an async resource, it should be properly implemented with SuspenseBoundary and use$

czoli1976:
@johnny-quesada-developer There are Two different things momoise and promisify, can do both in 1 go


Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

johnny-quesada-developer:
> this is not fixing the problem, the correct solution will be to properly migrate to OT.hasMediaProcessorSupport.promise... Since is an async resource, it should be properly implemented with SuspenseBoundary and use$

This is not fixing the underlying problem. The correct solution is to properly migrate to `OT.hasMediaProcessorSupport.promise`.

Since this is now an async resource, it should be implemented accordingly, potentially using a `SuspenseBoundary` and `use$`.

The current proposal is not the right solution, which is why I’m closing the task. I don’t want the PR review process to turn into designing the architecture of a feature/fix that was not part of the team’s commitment for the sprint.

That said, even if this change looks small (which is not), it needs to be properly planned. We are moving from a synchronous resource, which was very simple to consume, to an asynchronous one. We need to decide how that should be handled: whether we use Suspense as suggested, introduce an explicit loading state, or take another approach. I’ll need to spend some time evaluating what is both simpler and more appropriate.

The final solution should be accurate, easy to understand, and easy to maintain, especially since this is an open-source project.

PR #735, opened 2026-07-26
VIDSOL-1108: feat(settings): add WebRTC AEC/NS/AGC and Advanced Noise Suppression to Settings -> Audio

### LINKED ISSUE #734: Missing Audio Settings in Settings -> Audio
The Settings -> Audio tab is missing several audio options that should be configurable:

- Advanced Noise Suppression (Vonage Media Processor)
- WebRTC Echo Cancellation (AEC)
- WebRTC Noise Suppression (NS)
- WebRTC Auto Gain Control (AGC)

These settings are currently either not exposed or hidden behind device selection menus. They should be available in the Advanced Settings -> Audio panel for users to configure.

czoli1976:
## What

Adds missing audio processing options to the **Settings -> Audio** tab:

- **Advanced Noise Suppression** (Vonage Media Processor) — disabled when the browser doesn't support it
- **WebRTC Echo Cancellation** (AEC)
- **WebRTC Noise Suppression** (NS)
- **WebRTC Auto Gain Control** (AGC)

These settings were previously only available in device dropdown menus or not exposed at all.

## Changes

- Added , , , and  to 
- Added corresponding toggles in 
- Applied WebRTC audio constraints (, , ) in  when creating the publisher
- Migrated legacy  to  on restore
- Added  prop to  to support greyed-out toggles

Closes #734

## Testing

- Frontend lint + ts-check pass
- Manual testing: toggles visible in Settings -> Audio, disabled when media processor unsupported

## Note

This branch also includes Prettier/ESLint fixes from PR #721 that were needed to get the branch into a committable state.


VZaphod:
vcr:deploy

frontend/src/Context/AdvancedSettings/AdvancedSettings.tsx
```diff
@@ -183,4 +222,24 @@ function partialUpdate(partialState: Partial<advancedSettings>) {
...
 
+/**
+ * Whether the persisted settings already carry `advancedNoiseSuppressionEnabled`.
+ *
+ * The store merges the persisted payload over the initial state before handing it to the
+ * validator, so the restored object always has every key regardless of what was actually
+ * written. Reading the raw `{ s: state, v: version }` envelope is the only way to tell a
+ * pre-existing value apart from the default we just merged in.
+ */
+function isAdvancedNoiseSuppressionPersisted(): boolean {
```
VZaphod: @czoli1976 suggestion: instruct the AI to avoid refactoring the code, to keep the code simple and follow the existing guidelines. It adds extra code that we have to review and slows down our reviews.

Maybe AI thinks that has to follow Boy scouts rule but for these PRs won't help.

frontend/src/Context/AdvancedSettings/AdvancedSettings.tsx
```diff
@@ -68,28 +74,41 @@ const advancedSettingsSchema: z.ZodType<advancedSettings> = z.object({
...
-    key: 'advancedSettings',
+    key: ADVANCED_SETTINGS_STORAGE_KEY,
     selector: (state) =>
       ({
         ...state,
         isOpen: false,
       }) as advancedSettings,
 
     validator: ({ restored, initial }): advancedSettings => {
-      const restoredState = advancedSettingsSchema.safeParse(restored);
```
VZaphod, line 83: Please undo these refactors, leave only the part that can be copy pasted

frontend/src/Context/AdvancedSettings/AdvancedSettings.tsx
```diff
@@ -168,6 +187,26 @@ const advancedSettings$ = createGlobalState(INITIAL_STATE, {
         partialUpdate({ publisherStatisticsEnabled: value });
       };
     },
+    setAdvancedNoiseSuppressionEnabled(value: boolean) {
```
VZaphod, line 179: This is ok

frontend/src/locales/es-MX.json
```diff
@@ -63,6 +63,14 @@
   "advancedSettings.audio.publisherAudioFallback.description": "Cuando está habilitado, tu video deja de renderizarse en otros dispositivos durante condiciones de red deficientes para preservar el audio.",
   "advancedSettings.audio.subscriberAudioFallback.label": "Audio de respaldo del suscriptor",
   "advancedSettings.audio.subscriberAudioFallback.description": "Cuando está habilitado, recibes solo audio de otros participantes durante condiciones de red deficientes.",
+  "advancedSettings.audio.advancedNoiseSuppression.label": "Advanced Noise Suppression",
```
VZaphod: Please translate all languages

VZaphod:
@czoli1976 could you eliminate the Advances noise suppression toggle from the microphone menu?
Thanks!

frontend/src/Context/AdvancedSettings/AdvancedSettings.tsx
czoli1976: Done in 4d80c242. The validator is back to develop's shape — `fallbackState` restored, original if/return order, inline `'advancedSettings'` key, no cast on `safeParse`. The only change left in that block is the added migration branch.

frontend/src/locales/es-MX.json
czoli1976: Good catch — all four strings were English in de, it, es and es-MX, so `localeParity` passed while nothing was actually translated. Now translated in 4d80c242. Kept "WebRTC" and "Vonage Media Processor" as product terms.

frontend/src/Context/AdvancedSettings/AdvancedSettings.tsx
czoli1976: Fair, and taken on board — the refactors were unprompted tidying and I have reverted them. Guidance updated: no touching code the change does not require, no boy-scouting, keep to the existing patterns.

Two things also done in 4d80c242 from your other feedback:

- Advanced Noise Suppression removed from the in-call microphone menu now that it lives in Settings → Audio. Its `applyAudioFilter` / `clearAudioFilter` calls are reused in the settings handler, so it still applies to a running publisher instead of only at the next `initPublisher`.
- Descriptions now state the apply timing honestly. The three WebRTC constraints (`echoCancellation`, `noiseSuppression`, `autoGainControl`) are baked into the audio track at acquisition, so they say "Applies on your next call". Advanced Noise Suppression is a media processor filter and applies immediately, so it says so.

One thing worth flagging: the specs deleted in b9c4da70 and 881ee584 were what kept new-code coverage above the SonarCloud gate — it dropped to 77.7% against an 80% threshold. I have restored a much smaller set covering only the new code (store actions, the migration, `SwitchField`, and the new handler) rather than the originals.

czoli1976:
> @czoli1976 could you eliminate the Advances noise suppression toggle from the microphone menu? Thanks!

Your wish has been granted

VZaphod:
vcr:deploy

frontend/src/components/AdvancedSettingsDialog/components/AdvancedSettingsAudioTab/AdvancedSettingsAudioTab.spec.tsx
```diff
@@ -1,13 +1,42 @@
...
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { MemoryRouter } from 'react-router-dom';
+import type { Publisher } from '@vonage/client-sdk-video';
+import type { PublisherContextType } from '@Context/PublisherProvider';
+import usePublisherContext from '@hooks/usePublisherContext';
 import type { advancedSettings } from '@Context/AdvancedSettings';
 import advancedSettings$ from '@Context/AdvancedSettings';
 import AdvancedSettingsAudioTab from './AdvancedSettingsAudioTab';
 
+vi.mock('@hooks/usePublisherContext');
```
johnny-quesada-developer: Hi @czoli1976 ,

Please do not mock our own context. Mocking our code downgrades the quality of the test, we have this code convention both on **copilot-instructions.md** and **test-files.instructions.md**.  We cannot guaranty that the changes are working if the code of the provider is actually not running.

johnny-quesada-developer (changes requested):

frontend/src/Context/AdvancedSettings/AdvancedSettings.tsx
```diff
@@ -84,7 +94,17 @@ const advancedSettings$ = createGlobalState(INITIAL_STATE, {
       const fallbackState = initial as advancedSettings;
 
       if (restoredState.success) {
-        return restoredState.data;
+        if (isAdvancedNoiseSuppressionPersisted()) {
```
VZaphod: Non reverted code, please revert the changes in this function

frontend/src/Context/AdvancedSettings/AdvancedSettings.tsx
```diff
@@ -84,7 +94,17 @@ const advancedSettings$ = createGlobalState(INITIAL_STATE, {
...
+          return restoredState.data;
+        }
+
+        // First run after this setting shipped: adopt whatever the user already chose with the
+        // in-call Reduce Noise toggle, which persists under its own key.
+        return {
+          ...restoredState.data,
+          advancedNoiseSuppressionEnabled:
```
VZaphod: This should not be here, please remove

frontend/src/Context/AdvancedSettings/AdvancedSettings.tsx
czoli1976: Reverted in 71f541aa. `validator` is now byte-for-byte identical to `develop`:

```ts
if (restoredState.success) {
  return restoredState.data;
}
```

The migration branch is gone, and with it the `isAdvancedNoiseSuppressionPersisted()` helper and its `STORAGE_KEYS` / `tryCatch` imports, which existed only to serve it. `AdvancedSettings.tsx` now has no modified lines against `develop` at all — only additions: the four keys in `INITIAL_STATE`, the four in the schema, and the four setter actions.

One consequence worth stating before this merges, so it is your call and not a surprise: that branch was what carried a user's existing in-call **Reduce Noise** choice over to the new Settings → Audio toggle. Without it, anyone who had Reduce Noise on will find Advanced Noise Suppression **off** after upgrading, and the filter will stop applying until they switch it on again in the new location. It is a one-time, low-severity drop, and I am not arguing to keep the code — if you want the migration back, I would rather add it as a follow-up PR where it can be reviewed on its own merits than smuggle it into this one.

frontend/src/Context/AdvancedSettings/AdvancedSettings.tsx
czoli1976: Removed in 71f541aa — same block as above, gone in full.

frontend/src/components/AdvancedSettingsDialog/components/AdvancedSettingsAudioTab/AdvancedSettingsAudioTab.spec.tsx
czoli1976: Understood, and thanks for the pointer to the conventions — `vi.mock('@hooks/usePublisherContext')` was mocking our own hook, which is exactly what `test-files.instructions.md` rules out.

Fixed in 71f541aa. The spec now renders the real providers through `makeTestProvider`:

```tsx
makeTestProvider(
  [providers.advancedSettings, providers.runtime, providers.user, providers.session, providers.publisher],
  { advancedSettingsContext, publisherContext: { initialValue: { publisher, isPublishing: true } } }
)
```

So `PublisherProvider` and `usePublisher` actually run, and state is prepared through provider options rather than by mocking the context. The only thing still mocked is `@vonage/client-sdk-video` (automocked, the way `usePublisher.spec.tsx` does it) — a third-party SDK, which the convention allows. Setting `audioBitrateMode` also goes through `advancedSettingsContext.dialogState` now instead of poking the store directly, so it restores itself between tests.

frontend/src/components/AdvancedSettingsDialog/components/AdvancedSettingsAudioTab/AdvancedSettingsAudioTab.tsx
```diff
@@ -43,6 +62,23 @@ const AdvancedSettingsAudioTab = (): ReactElement => {
...
+  const handleAdvancedNoiseSuppressionChange = async (checked: boolean) => {
+    setAdvancedNoiseSuppressionEnabled(checked);
+
+    if (checked) {
+      await publisher?.applyAudioFilter({ type: 'advancedNoiseSuppression' });
+      return;
+    }
+
+    await publisher?.clearAudioFilter();
+  };
```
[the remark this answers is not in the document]
czoli1976: Good catch — fixed in 71f541aa. The handler now reverts the store when the SDK call rejects, so the toggle cannot claim a filter that is not running:

```tsx
const { error } = await tryCatch(async () => {
  if (checked) return publisher?.applyAudioFilter({ type: 'advancedNoiseSuppression' });

  return publisher?.clearAudioFilter();
});

if (!error) return;

console.error('AdvancedSettingsAudioTab: failed to apply advanced noise suppression', error);

setAdvancedNoiseSuppressionEnabled(!checked);
```

Covered by a new test that rejects `applyAudioFilter` and asserts both the store and the rendered switch fall back to off.

frontend/src/components/MeetingRoom/ReduceNoiseTestSpeakers/ReduceNoiseTestSpeakers.tsx
```diff
@@ -1,57 +1,24 @@
...
  * ReduceNoiseTestSpeakers Component
  *
- * This component displays options to enable advanced noise suppression
- * and to test the speakers.
- * @returns {ReactElement | false} Returns ReduceNoiseTestSpeakers component or false if the Vonage Media Processor is not supported.
+ * This component displays the option to test the speakers. Advanced noise suppression now lives in
+ * Settings -> Audio.
+ * @returns {ReactElement | false} Returns ReduceNoiseTestSpeakers component.
  */
 const ReduceNoiseTestSpeakers = (): ReactElement | false => {
```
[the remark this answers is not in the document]
czoli1976: Fixed in 71f541aa — signature and JSDoc are both `ReactElement` now. The component always renders; the `hasSpeakerDevices` check only gates the `SoundTest` child.

frontend/src/components/AdvancedSettingsDialog/components/AdvancedSettingsAudioTab/AdvancedSettingsAudioTab.tsx
```diff
@@ -43,6 +62,23 @@ const AdvancedSettingsAudioTab = (): ReactElement => {
...
+   * Advanced noise suppression is a media processor filter, so unlike the WebRTC constraints below
+   * it can be applied to a running publisher. Same calls the in-call menu used before this setting
+   * moved here.
+   * @param {boolean} checked - the requested state
+   */
+  const handleAdvancedNoiseSuppressionChange = async (checked: boolean) => {
+    setAdvancedNoiseSuppressionEnabled(checked);
+
+    if (checked) {
+      await publisher?.applyAudioFilter({ type: 'advancedNoiseSuppression' });
```
johnny-quesada-developer: Hi @czoli1976 ,

Instead of having spread through a lot of different places how or when to update the publisher we should move all advance settings publishers updates to one unique hook **frontend/src/Context/AdvancedSettings/AdvancedSettings/useAdvancesSettingsHandlers**:

We should move to **useAdvancesSettingsHandlers** the hook **useAdvancedSettingsVideoHandlers** plus the new **handleAdvancedNoiseSuppressionChange**. And delete the **useApplyAdvancedSettings**.

**useApplyAdvancedSettings** is not doing anything since **useAdvancedSettingsVideoHandlers** already updates the publisher. We should consistently update this settings only through this handlers.

frontend/src/components/AdvancedSettingsDialog/components/AdvancedSettingsAudioTab/AdvancedSettingsAudioTab.spec.tsx
```diff
@@ -27,7 +46,9 @@ describe('AdvancedSettingsAudioTab', () => {
   });
 
   it('renders the custom audio bitrate slider when custom mode is selected', () => {
-    render(<AdvancedSettingsAudioTab />, { dialogState: { audioBitrateMode: 'custom' } });
+    render(<AdvancedSettingsAudioTab />, {
+      advancedSettingsContext: { dialogState: { audioBitrateMode: 'custom' } },
```
johnny-quesada-developer: This is wrong, advances settings store is global, it doesn't use context. We don't need it on **makeTestProvider**.

This is the only test that uses this wrong called provider, lets delete **makeAdvancedSettingsProviderWrapper**, and make the corresponding fixes.

For your test you only need to add:
```tsx
beforeEach(() => {
    advancedSettings$.reset();
    ....
```

Plus the state preparation on your test:
```tsx
  it('renders the custom audio bitrate slider when custom mode is selected', () => {
    advancedSettings$.actions.setAudioBitrateMode('custom');

    render(<AdvancedSettingsAudioTab />);
    ...
```

We use **makeTestProvider** for actual **context/provider**.

frontend/src/Context/AdvancedSettings/AdvancedSettings.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+const loadStore = async () => {
+  const module = await import('./AdvancedSettings');
+  return module.default;
+};
+
+describe('advancedSettings$', () => {
+  beforeEach(() => {
+    window.localStorage.clear();
+    vi.resetModules();
```
johnny-quesada-developer: I'm almost sure **resetModules** is not necessary here.

frontend/src/Context/AdvancedSettings/AdvancedSettings.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  afterEach(() => {
+    window.localStorage.clear();
```
johnny-quesada-developer: This is not necessary, other test's should not expect the previous test to do cleanup, that's why we call `    window.localStorage.clear();`

frontend/src/Context/AdvancedSettings/AdvancedSettings.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  describe('audio processing actions', () => {
```
johnny-quesada-developer: **The test is not doing what the description says:**

You didn't setup an initial state so this code incapable of proving that there changes were applied correctly.  `validator` and `selector` logic are not been validated neither.

This store probably has coverage as side effect of the frontend/src/Context/AdvancedSettings/AdvancedSettings.spec.ts so if the test is not adding any value can be deleted.

libs/ui/src/components/SwitchField/SwitchField.tsx
```diff
@@ -21,14 +22,22 @@ const SwitchField = ({
   size,
   labelClassName,
   labelStyle,
+  disabled,
```
johnny-quesada-developer: You could potentially be shadowing and ignoring parameters that the actual **Field.Input** receives... **SwitchFieldProps** should be and extension of **FieldInputProps** plus whatever extra parameter we need to implement on **SwitchField**.

But honestly it just looks like **SwitchField** should not exists, the idea of **Field** using **composition** is to avoid having over complicated configuration based renders like this **SwitchField**.  With composition each segment remain 100% configurable and accesible and we wont need to fill the component with a lot of null validations or properties like **labelClassName**, and **labelStyle**

johnny-quesada-developer (changes requested):

frontend/src/components/AdvancedSettingsDialog/components/AdvancedSettingsAudioTab/AdvancedSettingsAudioTab.spec.tsx
czoli1976: You are right, thanks — `advancedSettings$` is a `createGlobalState`, not a context, and I should not have routed it through `makeTestProvider`. I took its presence in the `providers` enum as a signal it belonged there.

Fixed in 4cb88523, exactly as you described:

```tsx
beforeEach(() => {
  advancedSettings$.reset();
  ...
});

it('renders the custom audio bitrate slider when custom mode is selected', () => {
  advancedSettings$.actions.setAudioBitrateMode('custom');

  render(<AdvancedSettingsAudioTab />);
```

`makeAdvancedSettingsProviderWrapper` is deleted, along with its entry in the `providers` enum, its options/context types, its dependency row and its `switch` case in `makeTestProvider`.

One correction worth flagging: it was **not** the only test using it. `AdvancedSettingsStatisticsTab.spec.tsx` (already on `develop`, untouched by this PR) also listed `providers.advancedSettings` and plumbed an `advancedSettingsContext` option — though no test in that file ever passed one, so it was dead plumbing and removing it changes no behaviour. I had to touch that file to complete the deletion.

frontend/src/Context/AdvancedSettings/AdvancedSettings.spec.ts
czoli1976: Agreed on all three — the test set up no initial state, so it exercised neither `validator` nor `selector`, and `vi.resetModules()` plus the `afterEach` were both carrying their weight only because of that.

Deleted the file in 4cb88523 rather than inflating it. The four setter actions are exercised for real through `AdvancedSettingsAudioTab.spec.tsx`, which clicks the actual switches and asserts the store. If a genuine `validator` test is wanted (seeding `localStorage` and asserting what is restored), I would rather add it as its own focused spec than keep this one.

Related finding while moving the handlers: `advancedSettings$.reset()` in `afterEach` does **not** clear the persisted value, so `bitrateMode` leaked between tests in `useAdvancesSettingsHandlers.spec.ts` once the handlers started reading it from the store instead of from an argument. I applied your rule there too — the "when mode is not custom" test now sets `setBitrateMode('default')` explicitly instead of trusting the default.

frontend/src/components/AdvancedSettingsDialog/components/AdvancedSettingsAudioTab/AdvancedSettingsAudioTab.tsx
czoli1976: Done in 4cb88523. New `frontend/src/Context/AdvancedSettings/useAdvancesSettingsHandlers/` owns every advanced setting that has to reach the running publisher — the four video handlers plus `handleAdvancedNoiseSuppressionChange`. `useAdvancedSettingsVideoHandlers` is deleted and both dialog tabs now call the one hook.

Two notes on the shape:

- The handlers read `bitrateMode`/`customVideoBitrate` from the store rather than taking them as arguments. Otherwise the audio tab would have to select video state purely to satisfy the signature.
- The audio handler now resolves the publisher the same way the video handlers do (`meetingRoomPublisher ?? previewPublisher`), so the noise-suppression toggle also reaches the preview publisher in the waiting room. Previously it only looked at `usePublisherContext`, so in the waiting room it updated the store and nothing else.

**I have not deleted `useApplyAdvancedSettings`**, because I do not think it is a no-op, and I would rather check with you than ship a silent regression.

It runs on publisher *creation*; the handlers only run when a user moves a control. `usePublisherOptions` does not cover the gap — it passes `frameRate`, but `resolution: env.PUBLISHER_MAX_RESOLUTION` (not the user's chosen resolution) and no bitrate at all:

```ts
frameRate,
resolution: env.PUBLISHER_MAX_RESOLUTION,
```

So with it deleted, a user who picks a resolution or a custom bitrate in the waiting room loses both when the meeting-room publisher is created — nothing re-applies them until the control is touched again.

Deleting it cleanly means first passing the stored `resolution` and bitrate through `usePublisherOptions`. Happy to do that, here or as a follow-up — tell me which you prefer.

libs/ui/src/components/SwitchField/SwitchField.tsx
czoli1976: Fair — fixed in 4cb88523. `SwitchFieldProps` now derives from `FieldInputProps` instead of re-declaring the input's props:

```tsx
type SwitchInputProps = Omit<Extract<FieldInputProps, { variant: 'switch' }>, 'variant'>;

export type SwitchFieldProps = SwitchInputProps & {
  label: string;
  description?: string;
  labelClassName?: string;
  labelStyle?: CSSProperties;
};

// ...
<Field.Input variant="switch" {...inputProps} />
```

Nothing is shadowed now — `disabled`, `size` and anything later added to the switch input flow straight through, and `disabled` stopped being a prop I had to add by hand.

On "`SwitchField` should not exist": I agree with the reasoning, but it is used well beyond this PR and replacing it with compositional `Field` usage is a `libs/ui` change touching every call site. That feels like it deserves its own PR rather than riding along here — especially as this one has already been asked twice to stay minimal. Happy to open it if you want it.

frontend/src/Context/AdvancedSettings/AdvancedSettings.spec.ts
czoli1976: You were right. Addressed in 4cb88523 — the whole file is deleted, so `vi.resetModules()` goes with it.

It only existed to let each test re-import the store after seeding `localStorage`, which is the same reason the test was not proving anything (your other comment). There is now no `resetModules` anywhere in `frontend/src`.

frontend/src/Context/AdvancedSettings/AdvancedSettings.spec.ts
czoli1976: Agreed, and addressed in 4cb88523 — the file is deleted, so the redundant `afterEach` is gone too.

Point taken beyond this file: a test must not depend on the previous one having cleaned up, which is what the `beforeEach` is for. I applied the same reasoning in `useAdvancesSettingsHandlers.spec.ts`, where a test was relying on `bitrateMode` still being the default rather than setting it itself.

johnny-quesada-developer (dismissed):
lgtm

johnny-quesada-developer (dismissed):

VZaphod (approved):
LGTM

OscarFava (approved):

rserebrennykov (approved):
lgtm

Issue #738, opened 2026-07-28
Bug: Statistics reports the *configured* frame rate, not the measured one — and screen shares show none


czoli1976:
## Summary

The Frame rate row in Settings → Statistics does not report what the publisher is actually sending.
It echoes the value the publisher was *asked* for. Because the camera's frame-rate setting always
has a value (default 30), the number looks plausible and is easy to trust — but a camera throttled
by CPU or bandwidth still displays `30 fps`. A publisher with no configured frame rate displays
nothing at all.

## Current behavior

`libs/core/src/hooks/usePublisherStats/usePublisherStats.ts:84`

```ts
const frameRate = fixedFrameRate ?? null;
```

`fixedFrameRate` is passed in from the store by
`AdvancedSettingsStatisticsTab/components/PublisherStatistics.tsx:21-26`:

```ts
const fixedFrameRate = advancedSettings$.use.select((state) => state.frameRate);
```

That is the Video-tab setting, not a measurement. Nothing in the hook ever reads a frame rate from
the publisher, even though it already calls `publisher.getStats()` a few lines above and uses the
result for packets, bytes and bandwidth.

Consequences:

- **Camera:** always shows the configured value. If the encoder drops to 12 fps under CPU pressure,
  the panel still reads `30 fps` — the opposite of what an observability panel is for.
- **Any publisher without a configured frame rate:** renders the `-` fallback. This is how it
  surfaces for a screen share, whose frame rate is left to the browser.

## Expected behavior

Report the frame rate actually being sent, falling back to the configured value only when no
measurement is available.

## Note for whoever picks this up: the docs disagree on where it lives

- The [Publisher reference](https://vonage.github.io/conversation-docs/video-js-reference/latest/Publisher.html)
  documents `video.frameRate` — *"current average video frame rate"*.
- The [client observability guide](https://tokbox.com/developer/guides/client-observability/js/)
  lists only `bytesSent`, `packetsLost`, `packetsSent` and `layers` on the video stats object, and
  puts the rate in `layers[].encodedFrameRate` — *"actual encoding frame rate for this layer"*.

The TypeScript typings agree with the reference (`PublisherVideoStats = OutgoingTrackStats &
{ frameRate: number; layers?: VideoLayerStats[] }`), but a publisher with scalable video enabled
reports per layer. Reading both is what makes this robust.

Also worth knowing: `getStats()` errors unless the publisher is connected to a session **and**
[...]

PR #740, opened 2026-07-28
VIDSOL-1176: fix(stats): report the measured frame rate, not the configured one (#738)

czoli1976:
#### What is this PR doing?

Fixes the Frame rate row in Settings → Statistics, which never reported what a publisher was
actually sending.

`usePublisherStats` echoed `fixedFrameRate` — the value the publisher was *asked* for — and never
read a frame rate from the SDK, even though it already calls `getStats()` a few lines above for
packets, bytes and bandwidth:

```ts
// libs/core/src/hooks/usePublisherStats/usePublisherStats.ts:84 (before)
const frameRate = fixedFrameRate ?? null;
```

Because the camera's frame-rate setting always has a value (default 30), the number looked
plausible and was easy to trust — but a camera throttled to 12 fps by CPU pressure still displayed
`30 fps`, which is the opposite of what an observability panel is for. A publisher with nothing
configured — a screen share, whose frame rate is left to the browser — displayed the `-` fallback
instead, which is how the bug was first noticed.

The value is now resolved in this order:

1. the highest `layers[].encodedFrameRate` — the measured per-encoding rate
2. `stats.video.frameRate`
3. `fixedFrameRate` — the configured value, as a last resort

**Why layers come first.** Vonage's own documentation disagrees on where the frame rate lives: the
[Publisher reference](https://vonage.github.io/conversation-docs/video-js-reference/latest/Publisher.html)
documents `video.frameRate` as the *"current average video frame rate"*, while the
[client observability guide](https://tokbox.com/developer/guides/client-observability/js/) lists
only bytes, packets and layers on the video stats object and puts the rate in
`layers[].encodedFrameRate`, *"actual encoding frame rate for this layer"*. The SDK settles it:
`stats.video.frameRate` is marked deprecated in favour of `layers[].encodedFrameRate`
(`dist/js/opentok.js`), so the layers are read first and the top-level value is kept as the
fallback for publishers that report no layers. `stats.video.frameRate` is also a single number for
a publisher that may be encoding several simulcast layers, so the highest layer is the honest
answer.

The layer read lives in a `readHighestLayerFrameRate` helper, mirroring `readHighestLayerResolution`
in #742 so both sibling PRs are laid out the same way.

Note that `0` is deliberately preserved rather than treated as missing: a stalled encoding really is
sending 0 fps, and showing `0 fps` is more useful than showing `-`.

Scope is deliberately small: `libs/core` only, no behaviour change anywhere else.

1. Join a call and open **Settings → Statistics**.
2. Confirm the Frame rate row shows a plausible measured value for the camera.
3. Throttle the encoder — CPU pressure, or set a low frame rate in **Settings → Video** — and
   confirm the reported value follows what is actually sent rather than restating the setting.

The screen-share half of the symptom is easiest to see with the content-share statistics work, which
is not in this PR; on `develop` the fix is observable through the camera publisher.

Six unit tests cover it. Four on the helper: no layers reported, the highest layer winning, layers
that report no frame rate being ignored (which is how idle encodings arrive), and `0` surviving as a
real value. Two through the hook: `0` fps not being treated as missing, and the layer-only stats
shape.

**Not covered:** behaviour under **VP9 SVC** was not testable in the environment where this was
found, so no claim is made about it.

Resolves [VIDSOL-1176](https://jira.vonage.com/browse/VIDSOL-1176)

#### Checklist
[ ] Resolves a `Known Issue`.
If yes, which issue? [#738](https://github.com/Vonage/vonage-video-react-app/issues/738)

### DIFF
--- a/libs/core/src/hooks/usePublisherStats/helpers/index.ts
+++ b/libs/core/src/hooks/usePublisherStats/helpers/index.ts
@@ -1,0 +1,1 @@
+export { default as readHighestLayerFrameRate } from './readHighestLayerFrameRate';
--- a/libs/core/src/hooks/usePublisherStats/helpers/readHighestLayerFrameRate.spec.ts
+++ b/libs/core/src/hooks/usePublisherStats/helpers/readHighestLayerFrameRate.spec.ts
@@ -1,0 +1,31 @@
+import { describe, expect, it } from 'vitest';
+import type { VideoLayerStats } from '@vonage/client-sdk-video';
+import readHighestLayerFrameRate from './readHighestLayerFrameRate';
+
+describe('readHighestLayerFrameRate', () => {
+  it('returns null when no layers are reported', () => {
+    expect(readHighestLayerFrameRate([])).toBeNull();
+    expect(readHighestLayerFrameRate(undefined)).toBeNull();
+  });
+
+  it('returns the highest encoded frame rate', () => {
+    const layers = [{ encodedFrameRate: 5 }, { encodedFrameRate: 24 }] as VideoLayerStats[];
+
+    expect(readHighestLayerFrameRate(layers)).toBe(24);
+  });
+
+  it('ignores layers that report no frame rate, which is how idle encodings arrive', () => {
+    const layers = [
+      { width: 320, height: 180, bitrate: 0 },
+      { width: 640, height: 360, encodedFrameRate: 30, bitrate: 800000 },
+    ] as unknown as VideoLayerStats[];
+
+    expect(readHighestLayerFrameRate(layers)).toBe(30);
+  });
+
+  it('keeps 0 fps, which is a real value for a stalled encoding', () => {
+    const layers = [{ encodedFrameRate: 0 }] as VideoLayerStats[];
+
+    expect(readHighestLayerFrameRate(layers)).toBe(0);
+  });
+});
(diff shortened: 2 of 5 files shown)


VZaphod:
vcr:deploy

VZaphod (dismissed):
LGTM

Hossein-Movahed (dismissed):

OscarFava (dismissed):

VZaphod:
I think that this one requires a yarn lint:fix right? 
cc: @OscarFava

VZaphod (dismissed):
LGTM

OscarFava (dismissed):

Hossein-Movahed (dismissed):

OscarFava:
Hello, can you use the same approach as https://github.com/Vonage/vonage-video-react-app/pull/742? We should use the same way of working in both PRs, please, create a helper for the new function "readHighestLayerFrameRate".

czoli1976:
Done — `readHighestLayerFrameRate` is now a helper in `libs/core/src/hooks/usePublisherStats/helpers/`, laid out exactly like `readHighestLayerResolution` in #742: arrow function with a default export, re-exported through `helpers/index.ts`, with its own spec file (29f56d0). No behaviour change.

Heads up that both PRs create `helpers/index.ts` with a single export line, so whichever lands second will need a one-line merge there.

I also corrected the description while I was in here — it described a four-step resolution order that no longer matched the code (it listed `stats.video.frameRate` first and mentioned a `getSettings().frameRate` step that was never implemented). The code reads the layers first, because `stats.video.frameRate` is deprecated in the SDK in favour of `layers[].encodedFrameRate`.

rserebrennykov (approved):
LGTM

czoli1976:
@Hossein-Movahed , @OscarFava we got approval from QA, guess we need yours again now to merge ...

OscarFava (approved):

Issue #741, opened 2026-07-28
Bug: Statistics reports the *captured* resolution, not the encoded one


czoli1976:
## Summary

The Resolution row in Settings → Statistics reports what the camera is **capturing**, not what is
actually being **encoded and sent**. When the encoder scales down under CPU or bandwidth pressure —
precisely the situation this panel exists to diagnose — the row keeps showing the capture size.

This is the same shape of defect as #738, in the same hook: a configured/captured value presented
where a measured one belongs.

## Current behavior

`libs/core/src/hooks/usePublisherStats/usePublisherStats.ts:86-88`

```ts
const width = publisher.videoWidth();
const height = publisher.videoHeight();
const resolution = isNil(width) || isNil(height) ? null : { width, height };
```

`videoWidth()` / `videoHeight()` report the captured video. They do account for a browser refusing
the requested resolution, but nothing about them tracks encoder adaptation.

Consequence: on a constrained connection the panel reads `1280x720` while `640x360` is on the wire.
The number looks plausible, which is what makes it misleading rather than merely absent.

## Expected behavior

Report the encoded dimensions, falling back to the captured ones only when no layer data is
available.

## The data is already there

The hook already calls `publisher.getStats()` a few lines above and uses the result for packets,
bytes and bandwidth. The encoded size is on the same object — from the SDK's own typings
(`@vonage/client-sdk-video` 2.34.2, `dist/js/opentok.d.ts:708`):

```ts
export type VideoLayerStats = {
  codec: string;
  width: number;
  height: number;
  encodedFrameRate: number;
  bitrate: number;
  totalBitrate: number;
  scalabilityMode?: string;
  qualityLimitationReason?: VideoQualityReason;
};

export type PublisherVideoStats = OutgoingTrackStats & {
  frameRate: number;
  layers?: VideoLayerStats[];
};
```

`layers[].width` / `.height` are the encoded dimensions — "Encoded dimensions" in the
[client observability guide](https://tokbox.com/developer/guides/client-observability/js/).
`layers` is optional, hence the fallback.

## How to fix

In `usePublisherStats.ts`, replace lines 86-88:

```diff
-      const width = publisher.videoWidth();
-      const height = publisher.videoHeight();
-      const resolution = isNil(width) || isNil(height) ? null : { width, height };
+      const capturedWidth = publisher.videoWidth();
+      const capturedHeight = publisher.videoHeight();
+      const capturedResolution =
[...]

PR #742, opened 2026-07-28
VIDSOL-000: fix(stats): report the encoded resolution, not the captured one

czoli1976:
#### What is this PR doing?

Fixes the Resolution row in Settings → Statistics, which reported what the camera was *capturing*
rather than what was actually being encoded and sent.

```ts
// libs/core/src/hooks/usePublisherStats/usePublisherStats.ts:86-88 (before)
const width = publisher.videoWidth();
const height = publisher.videoHeight();
const resolution = isNil(width) || isNil(height) ? null : { width, height };
```

`videoWidth()` / `videoHeight()` account for a browser refusing the requested resolution, but not
for the encoder scaling down under CPU or bandwidth pressure — which is precisely the case this
panel exists to diagnose. On a constrained connection the row read `1280x720` while `640x360` was
on the wire. A plausible-looking number is worse than a missing one.

The encoded size is already on the object the hook fetches. `getStats()` is called a few lines
above for packets, bytes and bandwidth, and `VideoLayerStats` carries the encoded dimensions
(`dist/js/opentok.d.ts:708`):

```ts
export type VideoLayerStats = {
  codec: string;
  width: number;
  height: number;
  encodedFrameRate: number;
  bitrate: number;
  totalBitrate: number;
  scalabilityMode?: string;
  qualityLimitationReason?: VideoQualityReason;
};
```

Those are now preferred, with the captured size kept as the fallback for publishers reporting no
layers. Layers are compared by **area** rather than trusting array order, so simulcast and
single-layer publishers behave the same.

The runtime `typeof` guards are deliberate despite `width`/`height` being typed as required
numbers: #738 established that these stats objects do not always match their typings in practice.

Scope is deliberately small: two files in `libs/core`, no behaviour change anywhere else. This is
the sibling of #740 (frame rate) in the same hook — same class of defect, kept as a separate PR so
each reverts cleanly.

1. Join a call and open **Settings → Statistics**.
2. Note the Resolution row against the capture size.
3. Constrain the encoder — reduce available bandwidth, or apply CPU load — and confirm the reported
   resolution follows the encoder down instead of staying pinned at the captured size.

Verified this way during development: reducing bandwidth dropped the reported resolution to the top
active layer while capture stayed at 720p. On `develop` the row does not move.

Two unit tests cover it: the largest encoded layer winning over disagreeing captured dimensions,
and the fallback when no layers are reported.

**Not covered:** behaviour under **VP9 SVC** was not testable in the environment where this was
found, so no claim is made about it. `VideoLayerStats.scalabilityMode` is exposed, so an SVC-aware
refinement is reachable through the public API without dropping to raw WebRTC if it proves
necessary.

#### Checklist
[ ] Resolves a `Known Issue`.
If yes, which issue? [#741](https://github.com/Vonage/vonage-video-react-app/issues/741)


libs/core/src/hooks/usePublisherStats/usePublisherStats.spec.ts
```diff
@@ -52,6 +52,121 @@ describe('usePublisherStats', () => {
     });
   });
 
+  describe('encoded resolution', () => {
+    it('prefers the largest encoded layer over the captured dimensions', async () => {
```
OscarFava, line 56: This test is enough, we don not want to over test it.

libs/core/src/hooks/usePublisherStats/usePublisherStats.ts
```diff
@@ -83,9 +83,19 @@ const usePublisherStats = <Selected = PublisherInspectorStatistics | null>({
 
       const frameRate = fixedFrameRate ?? null;
 
-      const width = publisher.videoWidth();
-      const height = publisher.videoHeight();
-      const resolution = isNil(width) || isNil(height) ? null : { width, height };
+      /**
```
OscarFava: remove this comment

libs/core/src/hooks/usePublisherStats/usePublisherStats.ts
```diff
@@ -149,6 +159,32 @@ type PreviousPublisherVideoSample = {
   timestamp: number;
 };
 
+/**
```
OscarFava: remove comment

libs/core/src/hooks/usePublisherStats/usePublisherStats.ts
```diff
@@ -149,6 +159,32 @@ type PreviousPublisherVideoSample = {
...
+  layers: VideoLayerStats[] | undefined
+): { width: number; height: number } | null {
+  const sizes = (layers ?? [])
+    .map((layer) => ({ width: layer.width, height: layer.height }))
+    .filter((size) => typeof size.width === 'number' && typeof size.height === 'number')
+    .filter((size) => size.width > 0 && size.height > 0);
+
+  if (!sizes.length) return null;
+
+  return sizes.reduce((largest, size) =>
```
OscarFava: Can you check this SonarQube comment?

OscarFava (commented):
Some minor comments.

libs/core/src/hooks/usePublisherStats/usePublisherStats.ts
```diff
@@ -149,6 +159,32 @@ type PreviousPublisherVideoSample = {
...
+ * required numbers: an encoding that is active but not transmitting arrives without them, and the
+ * SDK assigns both straight from the underlying stats with no default - it guards them the same way
+ * in its own sort.
+ * @param {VideoLayerStats[] | undefined} layers - the publisher's active encoding layers
+ * @returns {{ width: number; height: number } | null} the largest encoded size, or null when no
+ * layer reports usable dimensions
+ */
+function readHighestLayerResolution(
+  layers: VideoLayerStats[] | undefined
+): { width: number; height: number } | null {
```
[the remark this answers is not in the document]
czoli1976: Moved `readHighestLayerResolution` into its own file (`helpers/readHighestLayerResolution.ts`), matching the `helpers/` pattern already used by `useVideoStats`. Fixed in 7a6ec57c.

libs/core/src/hooks/usePublisherStats/usePublisherStats.spec.ts
czoli1976: Agreed — kept this one as the integration-level proof that the hook wires the helper's result into `resolution`, and moved the edge-case permutations (idle layers with no dimensions, no layer encoded yet, no layers reported) into a focused unit-test file for the extracted helper (`helpers/readHighestLayerResolution.spec.ts`) instead. Fixed in 7a6ec57c.

libs/core/src/hooks/usePublisherStats/usePublisherStats.ts
czoli1976: Removed. Fixed in 7a6ec57c.

libs/core/src/hooks/usePublisherStats/usePublisherStats.ts
czoli1976: Removed — the function moved out to `helpers/readHighestLayerResolution.ts` without carrying the comment over. Fixed in 7a6ec57c.

libs/core/src/hooks/usePublisherStats/usePublisherStats.ts
czoli1976: Good catch — it was `sizes.reduce()` with no initial value, which Sonar flags because it throws on an empty array (even though it's guarded by the `if (!sizes.length) return null` above). Gave it an explicit initial value (`sizes[0]`) in the new helper file. Fixed in 7a6ec57c.

OscarFava (dismissed):
LGTM!

Hossein-Movahed (dismissed):

czoli1976:
@copilot resolve conflicts

johnny-quesada-developer (dismissed):

Hossein-Movahed (dismissed):

OscarFava (approved):

rserebrennykov (approved):
LGTM

Issue #607, opened 2026-06-29
Statistics: content-share (screen share) stream has no stats — panel only reflects the camera


czoli1976:
## Summary

The video statistics only ever reflect the **camera** publisher. When you share content (screen share), the shared-content stream gets **no statistics** — you'd expect to see stats for the content share alongside the camera while sharing.

## Current behavior

- Stats are produced by `useVideoStats(publisher)` → `readPublisherStats(publisher)`, which reads a **single camera publisher**: `publisher.videoWidth()` / `publisher.videoHeight()` for resolution and the camera track's `getSettings().frameRate`.
- The UI surface is `VideoStatsOverlay` (resolution + frame rate).
- The screen-share/content stream is published by a **separate** publisher — `useScreenShare` returns `screensharingPublisher: Publisher | null` — and that publisher is **never passed into the stats hook**, so nothing reports the content stream's resolution/frame rate.
- (Aside: the overlay is currently rendered only in the pre-join preview via `VideoContainer`; there is no in-call stats display for either camera or content yet.)

## Expected behavior

- While content sharing is active, the statistics panel should also show stats for the **content (screen-share) stream** — at minimum resolution and frame rate — *in addition to* the camera stats.
- Each set should be clearly labelled (e.g. **Camera** vs. **Content / Screen share**) so it's obvious which stream a number belongs to.

## Proposed approach

- Pass the existing `screensharingPublisher` into `useVideoStats` (it already accepts any `Publisher`), so the content stream's resolution/frame rate are read the same way the camera's are.
- Render a per-stream section in the stats UI — Camera and Content — and only show the Content section when a screen share is active.
- Optionally surface richer per-stream stats when published to a session (bitrate, packet loss, jitter) via the session stats API, for both camera and content.

## Affected code

- `libs/core/src/hooks/useVideoStats/*` — camera-only today (`readPublisherStats.ts`)
- `frontend/src/components/WaitingRoom/VideoStatsOverlay/*`
- `frontend/src/hooks/useScreenShare.tsx` — exposes `screensharingPublisher` that should feed the stats

czoli1976:
cc @OscarFava @VZaphod — flagging this one for visibility/triage. When content sharing is active, the stats only ever reflect the camera publisher; the screen-share publisher (`useScreenShare` → `screensharingPublisher`) is never fed into `useVideoStats`, so the shared content has no stats.

PR #747, opened 2026-07-28
VIDSOL-1119: refactor(screenshare): expose the screen-share publisher through a context

czoli1976:
#### What is this PR doing?

Pure refactor, **no user-visible change**: the screen-share publisher moves out of a `useRef`
inside `useScreenShare` and behind a context, so it can be reached from outside the meeting-room
component tree.

Today the publisher lives at `useScreenShare.tsx` in a ref, and reaches its only consumer through
`useMeetingRoom` → `MeetingRoom` → `VideoTileCanvas`. Anything outside that chain — notably the
Settings dialog — cannot see it at all. It also reaches consumers only because `setIsSharingScreen`
happens to trigger a re-render, which is incidental rather than designed.

This adds a `ScreenShareProvider` mounted on the meeting-room route and a `useScreenShareContext`
hook. `useMeetingRoom` consumes the context instead of calling `useScreenShare` directly; its
return shape is unchanged, so `MeetingRoom` and `VideoTileCanvas` need no edits.

**Why it is worth landing on its own:** it unblocks two follow-ups that both need the screen-share
publisher from the Settings dialog — content-share statistics (#607) and the video content hint —
and it rewrites the `useScreenShare` / `useMeetingRoom` specs, which is risk better reviewed in
isolation than buried in a feature PR.

The provider is mounted on the meeting-room route only. The Settings dialog also renders in the
waiting room, so consumers must tolerate an absent publisher rather than hard-require it; the
context is empty there by design.

Everything here is invisible if it works. The point is that screen sharing is unchanged:

1. Join a room and **start a screen share** — the tile appears for you and for a second participant.
2. **Stop and restart it.** Stale-publisher bugs show up on the second cycle, not the first.
3. Have a second participant share while you are sharing, to exercise the "someone else is sharing"
   path, which reads the publisher.
4. **Open the waiting room.** The provider does not exist on that route, so this confirms the
   absent-publisher case degrades quietly instead of throwing.

Verified locally across all four.

Resolves [VIDSOL-1192](https://jira.vonage.com/browse/VIDSOL-1192)

#### Checklist
[ ] Resolves a `Known Issue`.
If yes, which issue? Unblocks [#607](https://github.com/Vonage/vonage-video-react-app/issues/607).


OscarFava:
@czoli1976 do you have a ticket to link it? **check-pull-request-title** job is failing

frontend/src/Context/ScreenShareProvider/ScreenShareProvider.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+  default: () => ({
+    isSharingScreen: true,
+    isEntireScreen: false,
+    screensharingPublisher,
+    screenshareVideoElement: undefined,
+    toggleShareScreen,
+  }),
+}));
+
+describe('ScreenShareProvider', () => {
```
OscarFava: Does this test add any value? You are checking the same params you have mocked.

OscarFava (commented):
LGTM. But, do we have good coverage removing the test? If not, go ahead with it, thanks.

czoli1976:
> @czoli1976 do you have a ticket to link it? **check-pull-request-title** job is failing

I do not

frontend/src/Context/ScreenShareProvider/ScreenShareProvider.spec.tsx
czoli1976: Fair point — fixed in 56ddb04d. Now asserting reference equality on the whole context value (proves the provider passes it through unchanged rather than cloning/dropping fields, which the old per-field checks didn't actually guarantee) and added a separate test that children render.

frontend/src/hooks/useScreenShareContext.tsx
```diff
@@ -0,0 +1,15 @@
...
+ * its controls. Outside of a ScreenShareProvider - the waiting room, where the Settings dialog is
+ * also mounted - this returns an empty context, so consumers must treat the publisher as optional.
+ * @returns {ScreenShareContextType} - The current context value for the ScreenShare Context.
+ */
+const useScreenShareContext = (): ScreenShareContextType => {
+  const context = useContext<ScreenShareContextType>(ScreenShareContext);
+  return context;
+};
+
+export default useScreenShareContext;
```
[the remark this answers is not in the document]
czoli1976: @copilot fix it

frontend/src/Context/ScreenShareProvider/ScreenShareProvider.spec.tsx
```diff
@@ -0,0 +1,51 @@
+import type { Publisher } from '@vonage/client-sdk-video';
+import { render, renderHook, screen } from '@testing-library/react';
+import type { ReactNode } from 'react';
+import { describe, expect, it, vi } from 'vitest';
+import useScreenShare, { UseScreenShareType } from '../../hooks/useScreenShare';
```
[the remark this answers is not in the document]
czoli1976: @copilot  fix it

frontend/src/Context/ScreenShareProvider/ScreenShareProvider.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+    expect(screen.getByText('child content')).toBeInTheDocument();
+  });
+
+  it('degrades to an empty context outside the provider, as in the waiting room', () => {
+    const { result } = renderHook(() => useScreenShareContext());
+
+    expect(result.current.screensharingPublisher).toBeUndefined();
+  });
+});
```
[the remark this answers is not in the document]
czoli1976: @copilot  fix it

frontend/src/Context/ScreenShareProvider/ScreenShareProvider.spec.tsx
johnny-quesada-developer: It’s not fixed, @czoli1976. The only thing that **ScreenShareProvider** does is share the result of **useScreenShare**. If the hook is mocked, we’re basically testing **Vitest**.

czoli1976: Removed that test and Kept the two tests that check  children render, and the context degrades safely (not throws) outside the provider.

johnny-quesada-developer (dismissed):

johnny-quesada-developer (dismissed):

OscarFava (dismissed):

frontend/src/hooks/useMeetingRoom.ts
```diff
@@ -61,13 +61,11 @@ const useMeetingRoom = () => {
...
 
-  const {
-    isSharingScreen,
-    isEntireScreen,
-    screensharingPublisher,
-    screenshareVideoElement,
-    toggleShareScreen,
-  } = useScreenShare();
+  const { toggleShareScreen } = screenShare$.use.actions();
+  const isSharingScreen = screenShare$.use.select((state) => state.isSharingScreen);
```
johnny-quesada-developer: we introduced granular state and the ability of accessing the screenShare$ without having to be subscribe to the state changes

czoli1976: So, do you want me to change it or still fine as it is?

johnny-quesada-developer (approved):

OscarFava (approved):

rserebrennykov (approved):
LGTM

PR #748, opened 2026-07-28
feat(settings): split Settings -> Video into Camera and Screen Sharing sections

czoli1976:
#### What is this PR doing?

Splits **Settings → Video** into two labelled sections, **Camera** and **Screen Sharing**, each
carrying the same icon the meeting toolbar uses for that concept.

The tab is currently a flat list of four controls — Bitrate, Codec, Frame rate, Resolution — under
a single `Video` heading. Every one of them applies to the **camera** publisher only. The
screen-share publisher is created separately in `useScreenShare` and honours none of them, and
nothing in the tab says so, so the settings read as though they govern all outgoing video.

- **Camera** keeps all four existing controls, unchanged. No behaviour change, no store changes.
- **Screen Sharing** is deliberately **empty for now** — a heading plus a description line. It is
  the container for the screen-share settings that follow in later PRs, and it is present from the
  start so the tab's shape does not shift under users as those land.
- Both sections are always visible, including in the waiting room and when no share is running,
  since these are configured ahead of time.

Adds a small `SettingsSection` component to `libs/ui` — `{ title, icon?, description?, children }`,
no outer margin, DOM passthrough on the root. No such titled-section component existed:
`AdvancedSettingsStatisticsGroup` is a `Collapsible` built around read-only stat lists and
`LabeledValueList` takes `{label, value}` strings; neither accepts form-field children.

The heading uses the same `h4` treatment as the existing "Subscribers" group. Note this is the
first leading semantic icon next to a label in this dialog — the only other `VividIcon` uses are
the close button and two chevrons — so it sets a small precedent.

Locale keys land in all five files in this commit, as `localeParity.spec.ts` requires.

1. **Settings → Video** — two headings with icons matching the toolbar's camera and screen-share
   buttons.
2. **Camera** holds Bitrate, Codec, Frame rate and Resolution.
3. **Screen Sharing** shows its heading and description with no controls. This is expected.
4. Change Frame rate or Resolution and confirm they still apply — the controls moved in the JSX and
   should behave exactly as before.
5. Check from **both** the waiting room and in-call; the dialog is mounted on both routes.

Verified locally in the browser on both routes.

#### Checklist
[ ] Resolves a `Known Issue`.


czoli1976:

Issue #626, opened 2026-06-30
Enhancement: expose the Video Stats Overlay (#396) via the Settings menu instead of only the SHOW_VIDEO_STATS env flag


czoli1976:
## Summary

The **Video Stats Overlay** added in [#396](https://github.com/Vonage/vonage-video-react-app/pull/396) (VIDSOL-623) can only be turned on via the **build-time env flag `SHOW_VIDEO_STATS`** (default `false`), so end users (and people doing live demos) can't enable it without rebuilding. It should be controllable from the **Settings menu** at runtime.

## Current state

- The overlay is gated solely by the env flag:
  ```tsx
  // frontend/src/components/WaitingRoom/VideoContainer/VideoContainer.tsx:88
  {env.SHOW_VIDEO_STATS && isVideoEnabled && !isVideoLoading && (
    <div className="absolute left-4 top-3 z-10"><VideoStatsOverlay /></div>
  )}
  ```
- `SHOW_VIDEO_STATS` defaults to `false` (`env.ts:112`). There is **no UI toggle** for it.
- It currently renders only in the **waiting room**.

## Proposed change

Add a **"Show video stats overlay"** toggle to the Settings menu — the natural home is **Settings → Statistics** (`AdvancedSettingsStatisticsTab`), next to the existing "Enable publisher statistics" switch.

- Persist the choice in `localStorage` (same pattern as the other settings — `STORAGE_KEYS` + `getStorageItem`/`setStorageItem`).
- Gate the overlay on **the setting OR the env flag**, so `SHOW_VIDEO_STATS` keeps working as a deploy-time default:
  ```tsx
  const showVideoStatsOverlay = env.SHOW_VIDEO_STATS || advancedSettings$.use.select(s => s.showVideoStatsOverlay);
  // ...
  {showVideoStatsOverlay && isVideoEnabled && !isVideoLoading && ( <VideoStatsOverlay /> )}
  ```
- Optional follow-up: also surface the overlay **in-meeting** (it's waiting-room-only today) so the Settings toggle is useful during a call too.

## Affected code

- `frontend/src/components/AdvancedSettingsDialog/components/AdvancedSettingsStatisticsTab/AdvancedSettingsStatisticsTab.tsx` — add the `SwitchField`
- `frontend/src/Context/AdvancedSettings/*` (`advancedSettings$`) + `frontend/src/utils/storage.ts` (new `STORAGE_KEYS` entry) — state + persistence
[...]

PR #750, opened 2026-07-29
feat(settings): add self-view mirroring toggle and expose the video stats overlay

czoli1976:
> **Stacked PR — draft until the first batch lands.** This builds on #748 (Camera / Screen Sharing
> sections). Until #748 and the rest of that batch merge, the diff below also shows their commits;
> it shrinks to just this change as they land. Marked draft for that reason, not because the change
> is unfinished — it is tested and green.

#### What is this PR doing?

Adds two toggles to **Settings → Video → Camera**.

**1. Mirror my video.** Self-view mirroring is currently unconditional and implemented three
different ways: a Tailwind `child:-scale-x-100` on the waiting-room preview, an imperative
`element.style.transform = 'scaleX(-1)'` on the in-call self tile, and an inline
`transform: 'scaleX(-1)'` on the background-effects preview. All three now follow one store value
through a shared `useSelfViewMirroring` hook, so every view of your own camera agrees and the
toggle takes effect immediately.

Default is **on**, preserving today's behaviour. This is self-view only — CSS on the local element.
Remote participants are unaffected and the screen-share preview is deliberately not mirrored.

**2. Show resolution and frame rate.** The waiting room already has a `VideoStatsOverlay` that
renders a `720p 30fps` badge over the preview, fed by `useVideoStats`. It shipped but was never
reachable: it is gated on `env.SHOW_VIDEO_STATS`, which defaults to `false` and is read in exactly
one place, so with no env var set it is invisible. That env value now seeds the new store field
instead of acting as an outer gate — deployments that set it keep the badge on by default, and
everyone else can switch it on from Settings.

Scope note: the overlay exists only on the waiting-room preview today. This PR exposes it there and
does not extend it to the in-call self tile, so the setting has no visible effect during a meeting.
Called out here so it reads as a decision rather than an oversight.

1. **Settings → Video → Camera → Mirror my video.** Toggle it and watch the waiting-room preview
   flip immediately, with no rejoin.
2. Join a call and confirm the in-call self tile agrees with the setting, and that a remote
   participant's view of you is unchanged either way.
3. Open **background effects** and confirm that preview agrees too — it was the third, separate
   implementation.
4. **Show resolution and frame rate** — the badge appears on the waiting-room preview.

#### Checklist
[ ] Resolves a `Known Issue`.
If yes, which issue? [#626](https://github.com/Vonage/vonage-video-react-app/issues/626)


czoli1976:

Issue #608, opened 2026-06-29
Settings: expose video contentHint (camera + screen share) and show it in stats — currently hardcoded to 'detail'


czoli1976:
## Summary

VERA's use of `contentHint` is very limited: it is set in exactly one place — the screen-share publisher — where it is **hardcoded to `'detail'`**. There is no way for a user to choose the content hint for either the **camera** or the **screen share**, and the hint a publisher is actually using is **not shown anywhere** in the stats. This issue proposes (1) exposing the content hint as a user setting in **Settings → Video**, and (2) surfacing the active content hint in **Settings → Stats**.

## Background — what `contentHint` does

`videoContentHint` maps to the WebRTC `MediaStreamTrack.contentHint` and tells the encoder how to trade off under constrained bandwidth:
- **`detail`** — preserve spatial detail / resolution / sharpness at the cost of frame rate (good for slides, documents, code, UI).
- **`motion`** — prioritize frame rate / smoothness over per-frame sharpness (good for sharing video/animation, and a natural fit for camera).
- **`text`** — optimized for crisp text / line edges.
- **`''` (default)** — encoder decides.

## Current behavior

- The **only** usage in the codebase is the screen-share publisher, hardcoded:
  - `frontend/src/hooks/useScreenShare.tsx` → `initPublisher(..., { videoSource: 'screen', videoContentHint: 'detail', ... })`
- The **camera** publisher sets **no** content hint at all (`usePublisher` / `usePublisherOptions`), so it uses the encoder default.
- A repo-wide search (`grep -rni contenthint`, excluding `node_modules`) returns only that line plus its test (`frontend/src/hooks/tests/useScreenShare.spec.tsx`).

## Problems

1. **Hardcoded `detail` for screen share** is wrong for some content — sharing a video/animation under `detail` will look choppy (the encoder drops frame rate to keep sharpness). A user sharing motion content has no way to switch to `motion`.
2. **No control for the camera hint** — the camera is motion content and could benefit from `motion`, but it's left to the encoder default with no user override.
3. **The active hint is invisible** — users debugging quality in Settings → Stats can't see which content hint a publisher is using, unlike resolution / frame rate / bitrate / codec which are already shown.

### 1. Settings → Video: add a "Content hint" control
- Add a `SelectField` to `AdvancedSettingsVideoTab` (alongside the existing Resolution / Frame rate / Bitrate / Codec fields) with options: **Auto / Motion / Detail / Text**.
[...]

PR #751, opened 2026-07-29
feat(settings): add Optimize for content hint to both publishers

czoli1976:
> **Stacked PR — draft.** Base is #750 (self-view mirroring), so the diff shows only this change.
> It also needs #747 (screen-share publisher context) for the live screen-share path. Retarget to
> `develop` once its parents land. Draft for that reason, not because the change is unfinished.

#### What is this PR doing?

Adds an **Optimize for** dropdown to both the Camera and Screen Sharing sections, mapping to the
track's `videoContentHint`.

| Label | Value | Camera | Screen Sharing |
|---|---|---|---|
| Automatic | `''` | ✓ default | ✓ |
| Smooth motion | `'motion'` | ✓ | ✓ |
| Sharp detail | `'detail'` | ✓ | ✓ default |
| Readable text | `'text'` | — | ✓ |

Labels are outcome-phrased rather than spec terms. *Readable text* is omitted from Camera: legal on
a webcam, but meaningless.

**Defaults preserve today's behaviour exactly.** A full sweep found precisely one content hint in
the codebase — `videoContentHint: 'detail'` on the screen-share publisher. The camera, preview and
background publishers pass none, so the browser default `''` applies. Hence Screen Sharing defaults
to *Sharp detail* and Camera to *Automatic*, and no existing call changes behaviour on upgrade.

**Applies live on both sides, no republish.** Unlike codec priority, `videoContentHint` can be set
on a running track via `Publisher.setVideoContentHint()`. The camera path goes through the existing
`applyAdvancedSettingsToPublisher` helpers alongside frame rate, resolution and bitrate, following
the established publisher-first, store-second contract that the existing specs enforce. The
screen-share path is passed at `initPublisher` and applied live through the context from #747 when
a share is running.

SDK caveat checked and ruled out: canvas-sourced publishers get `'detail'` forced when no hint is
passed. No publisher in this app uses a canvas `videoSource` — background effects go through
`videoFilter` — so it is not in play.

1. **Settings → Video → Camera → Optimize for.** Change it mid-call; it should take effect without
   rejoining.
2. Start a screen share and change **Screen Sharing → Optimize for** while sharing — also live.
3. Confirm the screen share still defaults to *Sharp detail*, matching today's hardcoded value.

#### Checklist
[ ] Resolves a `Known Issue`.
If yes, which issue? [#608](https://github.com/Vonage/vonage-video-react-app/issues/608)


czoli1976:

PR #743, opened 2026-07-28
VIDSOL-1217: fix(settings): make codec priority reorderable by keyboard and touch

czoli1976:
#### What is this PR doing?

Makes the codec priority list in **Settings → Video** reorderable by keyboard and on touch.

The list was HTML5 drag-and-drop only — no key handlers, no ARIA, no touch path:

```tsx
// AdvancedSettingsCodecPriorityField.tsx (before)
<li
  key={codec}
  draggable
  onDragStart={…}
  onDragEnd={handleDragEnd}
  onDragOver={…}
  onDrop={…}
>
```

Two consequences:

- **Keyboard users cannot reorder the codecs at all.** There is no focusable control and no key
  handler anywhere in the field — a **WCAG 2.1.1 Keyboard (Level A)** failure.
- **The control is inert on touch.** HTML5 drag events are not synthesised from touch input, so on
  phones and tablets — where this dialog renders full-screen — the setting can be read but never
  changed.

This adds a **Move up** / **Move down** button to each row, disabled at the ends of the list, each
with a translated `aria-label` naming the codec it acts on (*"Move VP9 up"*). Both delegate to the
existing pure `reorderCodecPriority` helper, so ordering behaviour is unchanged and there is no
second implementation to keep in sync:

```ts
const moveCodec = (codec: AdvancedSettingsVideoCodec, offset: -1 | 1) => {
  const targetCodec = codecPriority[codecPriority.indexOf(codec) + offset];

  if (!targetCodec) return;

  setCodecPriority(reorderCodecPriority({ codecPriority, draggedCodec: codec, targetCodec }));
};
```

Drag-and-drop is deliberately untouched, so mouse users lose nothing.

**One piece of scope that is not accessibility.** The field's `data-testid`s were hardcoded, so the
component could only ever be rendered once per page. This adds an `idPrefix` prop (defaulting to
today's value, so nothing changes) to make a second instance addressable. That is groundwork for a
screen-share codec section; it is included here because it touches the same lines as the testids
above and would otherwise conflict.

New keys land in all five locale files in this commit, so `localeParity.spec.ts` stays green.

#### How should this be manually tested?

1. **Settings → Video → Codec → Manual** to reveal the priority list.
2. **Keyboard only:** Tab to a Move up / Move down button and activate it with Enter or Space. The
   codec should change position and the numbering update. On `develop` there is nothing to Tab to.
3. **Touch:** on a phone, tablet, or device emulation, tap the buttons. On `develop` dragging does
   nothing at all.
[...]

PR #752, opened 2026-07-29
feat(settings): add screen-share codec, scalable sharing, frame rate, resolution and bitrate

czoli1976:
> **Stacked PR — draft.** Base is #751 (Optimize for), so the diff shows only this change. It also
> needs #743 for the codec field's `idPrefix` prop. Retarget to `develop` once its parents land.

#### What is this PR doing?

Fills in the **Screen Sharing** section, closing the parity gap with Camera. Everything here is
screen-share only; the camera is untouched.

**Codec priority for the share.** Mirrors the Camera control, but with a third option, because the
share should follow the camera until you say otherwise:

| Option | Behaviour |
|---|---|
| Same as camera *(default)* | resolves to whatever the camera's codec settings evaluate to |
| Automatic | `preferredVideoCodecs: 'automatic'`, independent of the camera |
| Manual | reveals the share's own priority list |

Verified against the SDK: `preferredVideoCodecs` is a plain `PublisherProperties` field and its
validation is generic — no screen-source guard — so it is legitimate on a `videoSource: 'screen'`
publisher. It is init-only, so it applies the next time you start sharing. This reuses the existing
codec field via the `idPrefix` prop added in #743, so the two instances stay uniquely addressable.

**Scalable screen sharing.** `PublisherProperties.scalableScreenshare` is set nowhere in the app
today. Enabling it sends multiple quality layers so viewers on different connections each get what
they can handle. Default **off**, preserving current behaviour; init-only, same as the codec.

**Frame rate, resolution and bitrate for the share.** The screen publisher currently receives none
of these. Each control therefore has an explicit *Browser default* option — unlike the Camera
selects, which have no such entry — and choosing it passes nothing, so today's behaviour is the
default. These reuse `applyFrameRate` / `applyResolution` / `applyBitrate`, which already accept
any `Publisher` and early-return unless a video track exists, which is the correct guard for a
share that is not running.

Consolidated into one PR because all of it touches the same files — the store, the Video tab,
`useScreenShare`, and the five locale files. Splitting it would mean rebasing the same hot files
twice for no review benefit.

1. **Settings → Video → Screen Sharing.** The section now has Bitrate, Codec, Frame rate,
   Resolution, Optimize for and Scalable screen sharing — matching the Camera order.
2. Leave everything at its default, start a share, and confirm it behaves exactly as on `develop`.
3. Set **Codec → Manual** with a different order, then **start a new share** — codec is init-only,
   so an in-progress share will not change.
4. Set **Frame rate → 5** and confirm a slide-style share looks right; these apply to the running
   share.
5. Toggle **Scalable screen sharing** on and start a share with two viewers on different
   connections.

#### Checklist
[ ] Resolves a `Known Issue`.


czoli1976:

PR #770, opened 2026-08-17
VIDSOL-1159: Japanese language

VZaphod:
#### What is this PR doing?
Adds Japanese language

#### How should this be manually tested?
Check in UI

#### What are the relevant tickets?
A maintainer will add this ticket number.

Resolves [VIDSOL-1159](https://jira.vonage.com/browse/VIDSOL-1159)

#### Checklist
[ ] Branch is based on `develop` (not `main`).
[ ] Resolves a `Known Issue`.
[ ] If yes, did you remove the item from the `docs/KNOWN_ISSUES.md`? 
[ ] Resolves an item reported in `Issues`.
If yes, which issue? [Issue Number](https://github.com/Vonage/vonage-video-react-app/issues/)?

PR #774, opened 2026-08-19
VIDSOL-1212: Refresh the manifest hash so common:hash:check passes on develop

czoli1976:
## What is this PR doing?

Regenerates `libs/common/manifest.json` so `common:hash:check` passes on `develop` again. One line, no source changes.

## Why

`develop` currently fails its own hash gate:

```
❌ @vonage/video-common hash check failed
   Expected (manifest.json): b5397829f4d6f88be54dc1cb84ec416b458ad0353464a2648e5ba5d5f26a5087
   Actual   (source):        02ee0981fca4a09b104571c34d84f5728752d21afdf97435f654533d0eb86d60
```

The gate landed in #737 (`libs/common/manifest.json` + `.github/workflows/video-common-integrity.yml`). Four commits later, #770 (VIDSOL-1159, Japanese language) added a line to `libs/common/src/schemas/Lang.schema.ts` — a file inside the hashed `src` directory — without running `yarn common:hash:update`. The `validate` job on #770 failed and the PR was merged anyway.

Two things kept it quiet since 2026-08-17:

- `video-common-integrity.yml` is path-filtered to `libs/common/**`, so no later PR re-runs it.
- The failure surfaces instead as a **pre-push hook error on unrelated feature branches**, because `.husky/pre-push` runs `common:hash:check` on every push. Anyone who rebases onto current `develop` gets blocked from pushing by a breakage that has nothing to do with their change, and the message tells them to run `common:hash:update` and commit the manifest — which would quietly bundle this fix into whatever PR they happened to be working on.

## How was this verified

```
yarn nx run common:hash:check   # before: fails with the hashes above
yarn nx run common:hash:update  # regenerates the manifest
yarn nx run common:hash:check   # after: passes
```

Run on a clean checkout of `develop` at c7608f1b with no other local changes, so the new hash covers exactly the source currently on `develop`.

## Note for reviewers

Worth considering as a follow-up: the gate can only be trusted if `validate` is a required check. As it stands it failed on #770 and did not block the merge, which is how `develop` ended up in this state.


OscarFava:
Can we fix the conflicts? And then, we will approve, thanks. Please use last develop code, thanks.

jorgesanmartin-vng (dismissed):
👍

czoli1976:
@copilot resolve the merge conflicts in this pull request

czoli1976:
@copilot  please check if this is still needed after the conflicts have been resolved, the problem may not be there anymore so you need to double check and confirm if this is still needed

czoli1976:
> > @copilot  please check if this is still needed after the conflicts have been resolved, the problem may not be there anymore so you need to double check and confirm if this is still needed
> 
> I double-checked after merging `develop`: this branch is now identical to `develop` (no file diffs in `origin/develop...HEAD`), and `libs/common/manifest.json` has the same hash on both. So this PR is no longer needed.

@OscarFava will close it, develop surperseeded this already apparentely

PR #776, opened 2026-08-21
VIDSOL-1153: add Okta introspection middleware

Hossein-Movahed:
#### What is this PR doing?
- Adds `oktaAuthMiddleware`, opt-in via `OKTA_AUTH_ENABLED` (no-op by default, preserves current behavior everywhere until explicitly turned on).
- Checks `Authorization: Bearer <token>` first (mobile clients); falls back to `req.session?.accessToken` if absent (web populated by a separate, not-yet-built login flow, [VIDSOL-860](https://jira.vonage.com/browse/VIDSOL-860)).
- Validates via Okta's introspection endpoint (`${OKTA_ISSUER_URL}/oauth2/v1/introspect`), 401s on missing/inactive/failed, attaches `req.user` and calls `next()` on success.
- Wired onto `POST /createSession` and `POST /joinSession` only, as plain Express middleware ahead of the tRPC `videoHandler` mount (not `videoHandler.use$` its `ctx` doesn't carry `req`/`res`).
- Docs: `docs/CONFIGURATION.md` (full reference + a caveat that the session-fallback path can't be exercised end-to-end until [VIDSOL-860](https://jira.vonage.com/browse/VIDSOL-860) ships), short pointer in `README.md`, new vars in `backend/.env.example`.

**Tests:** unit tests cover session-fallback, inactive-token, and introspection-failure. Flag-off no-op, missing-token, and valid-token-success are not yet covered flagging for reviewers rather than claiming full coverage.

**Out of scope** (tracked separately, [VIDSOL-860](https://jira.vonage.com/browse/VIDSOL-860)): the actual web login/redirect/callback flow that populates `req.session.accessToken`. Without it, `OKTA_AUTH_ENABLED=true` blocks all web traffic  documented, not silently swallowed.

Resolves [VIDSOL-1153](https://jira.vonage.com/browse/VIDSOL-1153)

#### Checklist
[ ] Resolves a `Known Issue`.


johnny-quesada-developer (changes requested):
Hi @Hossein-Movahed ,

The implementation was made specifically for Okta, but this is something that is fundamentally standard OAuth 2.0/OIDC token introspection. Okta should be treated as a configurable provider, not baked into environment names, middleware names, types, and endpoint paths. 

**The token validation is incomplete**. The implementation needs to validate that the token was actually issued by the expected authority and intended for this application, using the standard claims available from introspection. The present of the token is not enought.

The middleware is being applied at the wrong level. You are covering only two routes while keeping all the others insecure. Is there a reason for leaving unprotected the rest of the API?

Also, the current tests are insufficient: mocking req, res, and next does not validate how this actually behaves in an Express request lifecycle... 

cc: @VZaphod , @OscarFava

Hossein-Movahed:
Hi @johnny-quesada-developer, thanks for the review, addressed the naming and claim validation points in the latest commit (everything is now provider-agnostic: `tokenAuthMiddleware`, `AUTH_ENABLED`,`OIDC_ISSUER_URL`, `OIDC_CLIENT_ID`, and `client_id` claim validation added).

On the route scope: the ticket explicitly scopes this to `/v2/createSession` and `/v2/joinSession` that's not an oversight, it's the stated requirement. That said, I found something worth flagging while looking into this: `GET /session/:room` in `backend/routes/session.ts` is a live, separately-mounted legacy route that also mints a join token (`videoClient.joinSession(...)`), completely outside this middleware's coverage. If that route stays reachable, it's a full bypass of this feature once `AUTH_ENABLED=true` anywhere real.

Is that route being deprecated as part of the v2 migration, or does it need OIDC coverage too? Flagging @VZaphod  before changing scope unilaterally, if it needs covering, the ticket scope needs updating first.

Also added a supertest-based integration test (`backend/tests/tokenAuthMiddleware.test.ts`) hitting the real mounted routes through the actual Express app, on top of the existing unit tests ,  addresses the point about mocked req/res/next not exercising the real request lifecycle.

backend/helpers/config.ts
```diff
@@ -22,6 +22,19 @@ const loadConfig = (): Config => {
 
   const loggerVerbose = process.env.LOGGER_VERBOSE === 'true';
 
+  const authConfig: AuthConfig = ((): AuthConfig => {
+    if (process.env.AUTH_ENABLED !== 'true') return { authEnabled: false };
+
+    const oidcIssuerUrl = process.env.OIDC_ISSUER_URL ?? '';
```
johnny-quesada-developer: check that oidcIssuerUrl is a valid url

backend/middleware/authMiddleware/authMiddleware.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+jest.unstable_mockModule('axios', () => ({
+  default: { post: axiosPostMock },
+}));
+
+const { default: authMiddleware } = await import('./authMiddleware');
+
+describe('authMiddleware', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
```
johnny-quesada-developer: no needed, this is already happening for all test in all the project

backend/middleware/authMiddleware/authMiddleware.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    process.env.VIDEO_SERVICE_PROVIDER = 'opentok';
```
johnny-quesada-developer: This configuration should be consume from the **config.ts** ... the. config.ts is the one that will have the proper schema validation.

backend/middleware/authMiddleware/authMiddleware.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const axiosPostMock = jest.fn<() => Promise<{ data: IntrospectionResponse }>>();
+
+const CONFIGURED_CLIENT_ID = 'test-client-id';
+
+jest.unstable_mockModule('axios', () => ({
+  default: { post: axiosPostMock },
+}));
+
+const { default: authMiddleware } = await import('./authMiddleware');
```
johnny-quesada-developer: using the config.ts you can import statically, and mock the implementation of the config.ts instead of having to complicate the test with dynamic imports

backend/middleware/authMiddleware/authMiddleware.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  it('calls next() with a valid Bearer header', async () => {
+    axiosPostMock.mockResolvedValue({
+      data: { active: true, sub: 'user-1', client_id: CONFIGURED_CLIENT_ID },
+    });
+
+    const { req, res, next } = createRequestParameters({
+      headers: { authorization: 'Bearer valid-token' },
+    });
+
+    await authMiddleware(req, res, next);
```
johnny-quesada-developer: use **supertest** to test this.

backend/middleware/authMiddleware/authMiddleware.ts
```diff
@@ -0,0 +1,127 @@
+import axios from 'axios';
+import type { NextFunction, Request, Response } from 'express';
+import { makeInternalErrorHandler, makeUnauthorizedErrorHandler } from '@api-lib/errors';
+import { assertResult } from '@api-lib/executions';
+import loadConfig from '../../helpers/config';
+
+const BEARER_PREFIX = 'bearer ';
```
johnny-quesada-developer: these are project configuration, they cannot be hidden in a file. They should go on env.sh and they should be overridable with .env, I think that `bearer` could some times be upper case. Also this will probably be part of the base configuration json schema.

backend/middleware/authMiddleware/authMiddleware.ts
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Validates the caller's OIDC access token against the configured provider's
+ * introspection endpoint. Opt-in via AUTH_ENABLED a no-op otherwise,
+ * preserving current behavior for deployments not yet on OIDC auth.
+ */
+async function authMiddleware(req: Request, res: Response, next: NextFunction): Promise<void> {
+  // Checked directly (not via loadConfig) so the no-op path never depends on
+  // unrelated config (e.g. video provider credentials) being valid.
+  const isAuthEnabled = process.env.AUTH_ENABLED === 'true';
```
johnny-quesada-developer: you should use config.ts

backend/middleware/authMiddleware/authMiddleware.ts
```diff
@@ -0,0 +1,127 @@
...
+    // tenant has no Custom Authorization Server, so `client_id` (not `aud`) is the reliable
+    // signal — without this check, a valid token from a different app in the same org would
+    // pass. Assumes one shared client id across mobile/web; revisit if VIDSOL-860 registers
+    // a separate web client id.
+    const isTokenValidForThisApp =
+      introspectionResponse.data.active === true &&
+      introspectionResponse.data.client_id === oidcClientId;
+
+    if (!isTokenValidForThisApp) {
+      throw makeUnauthorizedErrorHandler(UNAUTHORIZED_MESSAGE)(
```
johnny-quesada-developer: do not repeat the reason for the error, we need to be able to identify the specific step that failed

backend/middleware/authMiddleware/authMiddleware.ts
```diff
@@ -0,0 +1,127 @@
...
+        new Error('Token inactive or issued for a different client_id')
+      );
+    }
+
+    (req as RequestWithTokenAuth).user =
+      introspectionResponse.data as ActiveTokenIntrospectionResponse;
```
johnny-quesada-developer: there should be schema validation for data.. to make sure it returns the necessary expected information

backend/middleware/authMiddleware/authMiddleware.ts
```diff
@@ -0,0 +1,127 @@
...
+    next();
+  } catch (error) {
+    next(error);
```
johnny-quesada-developer: add an internal error wrapper for unhandled rejections

backend/middleware/authMiddleware/authMiddleware.ts
```diff
@@ -0,0 +1,127 @@
...
+  const authorizationHeader = req.headers.authorization;
+
+  if (authorizationHeader?.toLowerCase().startsWith(BEARER_PREFIX)) {
+    return authorizationHeader.slice(BEARER_PREFIX.length).trim();
+  }
+
+  // Session fallback for the web BFF login flow (VIDSOL-860, not yet built). Dead code
+  // today: no session middleware is registered in server.ts, so req.session is always
+  // undefined until that ticket lands.
+  return (req as RequestWithTokenAuth).session?.accessToken;
```
johnny-quesada-developer: we are using VCR o deploy vera, In vcr the instance of the server is ephemeral, so we are not going to store backend sessions, we are using instead JWT which are self contained. So we probably dont need this, better to remove it and adjust the test

backend/tests/authMiddleware.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// This needs to be set before the server is imported
+// and the import of the startServer cannot happen inside describe
+process.env.VIDEO_SERVICE_PROVIDER = 'opentok';
+process.env.AUTH_ENABLED = 'true';
+process.env.OIDC_CLIENT_ID = 'test-client-id';
+process.env.OIDC_ISSUER_URL = 'https://example.com';
+const startServer = (await import('../server')).default;
+
+describe('authMiddleware, mounted on the real app', () => {
```
johnny-quesada-developer: auth middleware test is duplicated

backend/server.ts
```diff
@@ -35,6 +35,18 @@ app.use(bodyParser.json());
...
+
+const routesExcludedFromAuth = new Set(['/_/health']);
+
+app.use((req, res, next) => {
+  if (routesExcludedFromAuth.has(req.path)) {
+    next();
+    return;
+  }
+
+  void authMiddleware(req, res, next);
```
johnny-quesada-developer: make the authMiddleware a builder so you can past as parameter the exclusions.

docs/CONFIGURATION.md
```diff
@@ -70,7 +70,27 @@ Enables the in-call issue reporting tool to file tickets directly into Jira.
 | `JIRA_EPIC_LINK` | Epic link field value |
 | `JIRA_EPIC_URL` | URL to the target epic |
 
----
+#### OIDC token authentication (optional)
+
+Protects `POST /v2/createSession` and `POST /v2/joinSession` by validating OIDC access tokens through the configured OAuth 2.0 introspection endpoint.
```
johnny-quesada-developer: description outdated

docs/CONFIGURATION.md
```diff
@@ -70,7 +70,27 @@ Enables the in-call issue reporting tool to file tickets directly into Jira.
...
+| `OIDC_CLIENT_ID` | if enabled | OIDC application client ID. |
+| `OIDC_ISSUER_URL` | if enabled | Provider org root URL. Introspection uses `${OIDC_ISSUER_URL}/oauth2/v1/introspect`. |
+
+```ini
+AUTH_ENABLED='true'
+OIDC_CLIENT_ID='your-client-id'
+OIDC_ISSUER_URL='https://your-org.okta.com'
+```
+
+The middleware checks for a Bearer token first, then falls back to req.session.accessToken. Both use the same introspection flow. Requests return 401 when the token is missing, inactive, invalid for this application, or introspection fails.
```
johnny-quesada-developer: req.session.accessToken is not going to be there an the header for the token should be configurable, not always Authorization

README.md
```diff
@@ -118,6 +118,10 @@ This application provides features for common conferencing use cases, such as:
 
 See [Getting Started](./docs/GETTING_STARTED.md) for the full setup guide (Vonage account, environment variables, multi-device testing, and VCR deployment).
 
+### OIDC token authentication (optional)
+
+The backend can validate OIDC access tokens on `/v2/createSession` and `/v2/joinSession` via `AUTH_ENABLED` (opt-in, `false`/unset by default), `OIDC_CLIENT_ID`, and `OIDC_ISSUER_URL`. Mobile clients authenticate via `Authorization: Bearer <token>`; web clients via a server-side session (`req.session.accessToken`). See [Configuration](./docs/CONFIGURATION.md#oidc-token-authentication-optional-opt-in) for the full variable reference and current limitations.
```
johnny-quesada-developer: description is outdated... and both web and mobile they authenticate  the same

johnny-quesada-developer (changes requested):

backend/middleware/authMiddleware/authMiddleware.test.ts
Hossein-Movahed: Tried a fully static import here and it doesn't actually work, this repo's Jest is real ESM (`ts-jest` + `useESM: true`), not the Babel hoisted CJS most `jest.mock()` guides assume. Tested it two ways: one crashed with a TDZ error (static import evaluates the real module before the mock registers), the other ran but `loadConfig` came back unmocked. `logger.test.ts` only gets away with `jest.mock()` because it never statically imports the thing that needs config.ts either  it's dynamic under the hood too.

So I kept `unstable_mockModule` + dynamic import (same pattern as `feedback.test.ts/session.test.ts/loggerService.test.ts` it's the only thing that works for this setup), but fixed the actual complaint: it now mocks `config.ts` directly instead of the bespoke l`oadAuthConfig `helper, which is gone.

If you'd rather have a truly static import, only way there is DI-ing config into `authMiddleware()` instead of importing it,   but that'd be inconsistent with how everything else in backend/ does it. 
Lmk if you want that anyway.

OscarFava (dismissed):
LGTM! 
My only concern is how are the errors from middleware be shown in FE?

backend/helpers/config.test.ts
johnny-quesada-developer, line 1: this file only was moved.

backend/middleware/authMiddleware/authMiddleware.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  app.use(errorHandler);
+
+  return app;
+}
+
+describe('authMiddleware', () => {
+  const originalEnv = process.env;
+
+  beforeEach(() => {
+    process.env = {
```
johnny-quesada-developer, line 28: We override the source not the helper that loads the configuration, so the test are closer to reality... it also removes the need of dynamic imports

backend/middleware/authMiddleware/authMiddleware.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      ...originalEnv,
+      AUTH_ENABLED: 'true',
+      OIDC_ISSUER_URL: 'https://example.com',
+      OIDC_CLIENT_ID: CLIENT_ID,
+    };
+  });
+
+  afterEach(() => {
+    process.env = originalEnv;
```
johnny-quesada-developer, line 37: jest doesn't have stub

backend/tests/helpers/doPartialMock.ts
```diff
@@ -0,0 +1,43 @@
...
+ *   (actual) => ({
+ *     ...actual,
+ *     <overrides/spies go here>
+ *   })
+ * );
+ *
+ * // Test target module after mocking
+ * const { default: authMiddleware } = await import('./authMiddleware');
+ */
+async function doPartialMock<T extends Record<string, Any>>(
```
johnny-quesada-developer, line 28: Makes jest dynamic more linear an consistent... We can keep it as an utility, but at the end the test did't actually required dynamic imports as I originally thought.

```tsx
const { default: loadConfig } = await doPartialMock(
 *   '../../helpers/config',
 *   () => import('../../helpers/config'),
 *   (actual) => ({
 *     ...actual,
 *     <overrides/spies go here>
 *   })
 * );
```

backend/middleware/authMiddleware/authMiddleware.test.ts
johnny-quesada-developer: Hi @Hossein-Movahed, I made the changes... The file needed the dynamic render becase we were mocking our function instead of the environment variables . Now the test looks a lot cleaner.

Also while reviewing the test I notice varios problems like lack of isolation between one test and the other that there was a huge gap of proper data integrity validations. So example I was able to remove huge portions of mocked data and the test still where passing cause the underline new logic wasn't defensive enough. 

I also moved the AUTH config at project level instead of hardcode parameters:

```bash
# Auth OIDC configuration
export AUTH_HEADER_NAME='authorization'
export AUTH_SCHEME='Bearer'
export OIDC_INTROSPECT_PATH='/oauth2/v1/introspect'
export AUTH_INTROSPECTION_TIMEOUT_MS=5000
```

CC: @OscarFava

johnny-quesada-developer:
> Hi @johnny-quesada-developer, thanks for the review, addressed the naming and claim validation points in the latest commit (everything is now provider-agnostic: `tokenAuthMiddleware`, `AUTH_ENABLED`,`OIDC_ISSUER_URL`, `OIDC_CLIENT_ID`, and `client_id` claim validation added).
> 
> On the route scope: the ticket explicitly scopes this to `/v2/createSession` and `/v2/joinSession` that's not an oversight, it's the stated requirement. That said, I found something worth flagging while looking into this: `GET /session/:room` in `backend/routes/session.ts` is a live, separately-mounted legacy route that also mints a join token (`videoClient.joinSession(...)`), completely outside this middleware's coverage. If that route stays reachable, it's a full bypass of this feature once `AUTH_ENABLED=true` anywhere real.
> 
> Is that route being deprecated as part of the v2 migration, or does it need OIDC coverage too? Flagging @VZaphod before changing scope unilaterally, if it needs covering, the ticket scope needs updating first.
> 
> Also added a supertest-based integration test (`backend/tests/tokenAuthMiddleware.test.ts`) hitting the real mounted routes through the actual Express app, on top of the existing unit tests , addresses the point about mocked req/res/next not exercising the real request lifecycle.



> LGTM! My only concern is how are the errors from middleware be shown in FE?

It depends on the accept headers... if a json request is sent we return a JSON error and if HTML request the error handler returns and HTML Page with the error. In both cases the configure logger save the incidents. The errors we'll show are the human error fallbacks:
* Token introspection request to the identity provider failed
* Token introspection response failed schema validation

johnny-quesada-developer (approved):

OscarFava (approved):

backend/jest/setEnvVars.js
```diff
@@ -8,6 +8,7 @@ process.env.JIRA_EPIC_LINK = 'jiraEpicLink';
 process.env.JIRA_EPIC_URL = 'jiraEpicUrl';
 process.env.VONAGE_APP_ID = 'vonageAppId';
 process.env.VONAGE_PRIVATE_KEY = 'vonagePrivateKey';
+process.env.VIDEO_SERVICE_PROVIDER = 'opentok';
```
VZaphod, line 11: isn't the current provider vonage?

Hossein-Movahed: Yeah, production uses Vonage, `opentok` is just used in the test because it's the simpler config shape to set up. The middleware itself is provider agnostic either way 🙂

rserebrennykov (approved):
LGTM

PR #786, opened 2026-09-01
VIDSOL-860: add Okta Web BFF login flow

Hossein-Movahed:
#### What is this PR doing?

Adds the Okta Backend-for-Frontend (BFF) login flow for web, on top of VIDSOL-1153's Bearer-only `authMiddleware`. Without this, a browser had no way to obtain a token, so every request 401'd once `AUTH_ENABLED=true`.

- `GET /auth/signin` → `GET /api/auth/callback/okta`: Authorization Code + PKCE flow. The PKCE transaction (`state`/`code_verifier`) and the resulting access token are both stored server-side via the existing `SessionStorage` abstraction (`InMemorySessionStorage` locally, `VcrSessionStorage` on VCR) — the browser only ever holds opaque, `HttpOnly` cookies, never the real token.
- `authMiddleware` now checks the Bearer header first (Mobile), then falls back to the session cookie (Web) — same introspection call, one shared `OIDC_CLIENT_ID` app registration for both.
- Auth config validation consolidated into a single `AuthConfigSchema` (Zod discriminated union) shared between `loadConfig` and `authMiddleware`, replacing hand-rolled `if` checks with one schema-validated source of truth.
- Automatic redirect to `/auth/signin` on any 401 (`fetchWithAuthRedirect`), with `credentials: 'include'` / `withCredentials: true` added everywhere the frontend calls the backend — needed because cookies aren't port-scoped but `fetch`'s default credentials mode is same-origin only.
- `returnTo` is validated against open-redirect (must be a same-origin relative path) so the user lands back where they were instead of always `/`.
- `getCookieValue` extracted to `libs/common/src` (isomorphic, no Node APIs) since it's now shared between the callback handler and `authMiddleware`.
- Vite dev-server proxy added for `/auth/signin` and `/api/auth/callback/okta`, since Okta's registered local redirect URI is fixed to the frontend's own origin.
- `env.sh`, `vcr-gha.yml`, `.env.example`, `docs/CONFIGURATION.md`, and `README.md` updated for the new flow and env vars.

1. Set `AUTH_ENABLED=true` plus `OIDC_CLIENT_ID`, `OIDC_ISSUER_URL`, `OIDC_WEB_REDIRECT_URI` (DEV values already default in `env.sh`).
2. `yarn dev`, open the app in an incognito window, and trigger a protected action (e.g. create a room).
3. Confirm you're redirected to Okta, log in, and land back on the page you started from (not always `/`).
4. Confirm the protected action now succeeds, and that only an opaque `oidc_session_id` cookie is visible in devtools — never the real access token. 

Resolves [VIDSOL-860](https://jira.vonage.com/browse/VIDSOL-860)

#### Checklist
[ ] Resolves a `Known Issue`.


Hossein-Movahed:
vcr:deploy

OscarFava (dismissed):

backend/routes/auth/helpers/isSafeReturnToPath.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Guards against open-redirect: only a same-origin relative path is a safe `returnTo` —
+ * `//evil.com`, `/\evil.com`, and absolute URLs all resolve to another host in a browser.
```
VZaphod, line 3: maybe remove the comment?

Hossein-Movahed: Removed

VZaphod (approved):
LGTM

Issue #789, opened 2026-09-02
Safari: background replacement kills the WebContent process after ~90-180s (page goes blank, camera permission re-prompts)


czoli1976:
## Summary

On Safari, applying a **background replacement** effect terminates the WebContent process after roughly 90–180 seconds at 720p. The page goes blank and Safari re-prompts for camera permission, because the tab is now running in a fresh web process.

**Background blur is unaffected.**

Reproducible in the Waiting Room on `meet.vonagenetworks.net` and locally on `develop`.

## Steps to reproduce

1. Open the Waiting Room in Safari (26.6.2 confirmed).
2. Open **Video effects** and select any background replacement image.
3. Apply, and leave the tab focused.
4. Within ~90–180 seconds the page goes blank and the camera permission prompt returns.

## Root cause

This is a bug in `@vonage/client-sdk-video`, not in VERA. The bare-SDK reproduction below has no React and no VERA code and still fails.

The media processor picks a rendering backend per filter type, defaulting to `CANVAS`:

```js
// getDefaultConfig
renderingOptions: { type: RenderingType.CANVAS }
```

`BackgroundBlurConfig.getConfig` overrides that on Safari:

```js
if (env.isSafari) {
  config.renderingOptions.type = RenderingType.WEBGL;
}
```

`BackgroundReplacementConfig.getConfig` has **no such branch**, so replacement stays on `CANVAS` and composites 720p frames on a 2D canvas until WebKit reclaims the process.

Each SDK bundle contains exactly one `isSafari -> WEBGL` override, and it is in the blur path. The Safari workaround was written for blur and never applied to replacement.

**`2.36.0-alpha.35` does not fix this** — `BackgroundReplacementConfig` is unchanged, the default is still `CANVAS`, and the single override is still in the blur path. Same segmenter model (6.0.0) and mediapipe (0.10.20). Upgrading is not a way out.

## Evidence

Eleven automated runs via `safaridriver` on Safari 26.6.2. "Died" means the WebDriver session went invalid, which is what happens when the web process is replaced.

| Context | Effect | Rendering | Resolution | Result |
| --- | --- | --- | --- | --- |
| VERA | none | – | 720p | survived 5m |
| VERA | blur | WEBGL | 720p | survived 5m |
| bare SDK | blur | WEBGL | 720p | survived 5m |
| VERA | replacement | CANVAS | 720p | died 90s |
| VERA | replacement | CANVAS | 720p | died 97s |
[...]

PR #790, opened 2026-09-02
VIDSOL-1213: fix(effects): keep Safari background replacement on the WebGL backend

czoli1976:
#### What is this PR doing?

Fixes #789 — on Safari, applying a background replacement effect kills the WebContent process after 90–180 seconds. The page goes blank and Safari re-prompts for camera permission, because the tab is now in a fresh web process. Background blur is unaffected.

The media processor picks a rendering backend per filter type and defaults to `CANVAS`. `BackgroundBlurConfig.getConfig` overrides that to `WEBGL` when the browser is Safari; `BackgroundReplacementConfig.getConfig` has no such branch, so replacement composites 720p frames on a 2D canvas until WebKit reclaims the process.

**This is an SDK bug, not a VERA bug** — a bare page using only `OT.initPublisher` and `applyVideoFilter`, with no React and no VERA code, fails the same way. `2.36.0-alpha.35` does not fix it: `BackgroundReplacementConfig` is unchanged there. #789 carries the SDK-side diff to hand to Vonage.

Until the SDK ships a fix, the backend can be pinned from our side. The SDK merges caller options over its own defaults, so passing `renderingOptions` through the filter reaches exactly the configuration the missing override would have set:

```ts
{ type: 'backgroundReplacement', backgroundImgUrl: url, renderingOptions: { type: 'WEBGL' } }
```

A video filter reaches the SDK from four places, so a single helper is applied at each:

- `applyBackgroundFilter` — runtime `applyVideoFilter`
- `usePreviewPublisher` — initial `videoFilter`
- `usePublisherOptions` — initial `videoFilter`
- `useBackgroundPublisher` — initial `videoFilter`

The helper reuses the existing `isWebKit()` from `@web/platform`, and narrows to `backgroundReplacement` only, so blur and the no-filter path are untouched. In `applyBackgroundFilter` the override is applied at the SDK call only, so the value written to storage and to React state stays clean.

`renderingOptions` is not on the SDK's `BackgroundReplacementFilter` type, so the helper declares the extra property locally rather than casting.

Safari is required — Chrome and Firefox do not exhibit this.

1. Open the Waiting Room in Safari, open **Video effects**, pick any background image, Apply.
2. Leave the tab focused for 5 minutes. It should keep running: no blank page, no repeated camera permission prompt.
3. On `develop` the same steps blank the page within ~90–180 seconds.
4. Repeat with **blur** and with **no effect** — both were already fine and should stay fine.
5. Repeat in the meeting room, and after a reload with a replacement background already stored, to cover the `initPublisher` paths.

#### Verification already done

Automated under `safaridriver` on Safari 26.6.2, driving the real Waiting Room:

| Build | Scenario | Result |
| --- | --- | --- |
| develop | waiting room + replacement | died 90s |
| develop | waiting room + replacement | died 97s |
| this branch | waiting room + replacement | **survived 6m** |

Supporting runs isolating the cause, from #789:

| Context | Effect | Rendering | Resolution | Result |
| --- | --- | --- | --- | --- |
| bare SDK | replacement | CANVAS | 720p | died 141s |
| bare SDK | replacement | CANVAS | 480p | survived 5m |
| bare SDK | replacement, forced | WEBGL | 720p | survived 5m |
| bare SDK | replacement, patched SDK | WEBGL | 720p | survived 5m |

Holding page, image, resolution and SDK version constant and changing only the rendering backend turns a reliable failure into a clean run.

`ts-check`, `lint` and the affected unit suites pass locally.

#### Caveats

- Reproduced with a preview publisher; not yet reproduced or verified inside a published session.
- Chrome and Firefox assumed unaffected, not re-tested here.
- Not measured whether the WebGL backend changes segmentation quality or CPU cost.
- This is a workaround. It should be removed once the SDK gives `BackgroundReplacementConfig` the branch blur already has.

#### Checklist
[ ] Resolves a `Known Issue`.
If yes, which issue? [#789](https://github.com/Vonage/vonage-video-react-app/issues/789)


czoli1976:
vcr:deploy

czoli1976:
@copilot Review this PR, apply all needed changes and solve any conflict  and rebase (only if needed)  then  and Explain Very Clearly the Value for @VZaphod, Tag him to ask for next move on this

czoli1976:
@VZaphod do you want to adop this workarround or wait for the Client SDK to address it ?

czoli1976:
Closing as the Team prefers to wait for the Client SDK to fix the FF and Safari Issues instead.

